PR #74, opened 2026-04-08
feat: add WebSocket client for backend print dispatch

Knorcedger:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&task=69fd2d528ccea86c7add5b53)

## Summary
- New `wsClient.ts` that connects to the backend WebSocket server
- Registers as a printer server with venueId
- Receives `printRaw` messages with base64-encoded ESC/POS data
- Sends raw bytes to printers via TCP socket
- Auto-reconnect with exponential backoff
- Runs alongside existing Express server (backward compatible)

Part of the printer system refactor: FE -> GQL -> BE -> WS -> **PrinterServer** -> TCP -> Printer

## Test plan
- [ ] Printer server connects to backend WebSocket on startup
- [ ] Receives printRaw message and sends data to printer via TCP
- [ ] Auto-reconnects if WebSocket drops
- [ ] Existing HTTP endpoints still work

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/modules/wsClient.ts
```diff
@@ -0,0 +1,168 @@
+/**
+ * WebSocket client that connects to the Quickord backend.
+ * Receives raw ESC/POS print commands and sends them to local printers via TCP/serial.
```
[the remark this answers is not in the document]
Knorcedger: Fixed — updated comment to TCP-only.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,168 @@
...
+
+async function sendToPrinter(
+  ip: string,
+  port: number,
+  data: Buffer
+): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const socket = net.connect(
+      { host: ip, port: port || 9100, timeout: 5000 },
+      () => {
```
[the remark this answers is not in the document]
Knorcedger: Fixed — using `socket.setTimeout(SOCKET_TIMEOUT)` after connect.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,168 @@
...
+      }
+    );
+    socket.on('error', (err) => {
+      socket.destroy();
+      reject(err);
+    });
+    socket.on('timeout', () => {
+      socket.destroy();
+      reject(new Error('Connection timeout'));
+    });
```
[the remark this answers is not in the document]
Knorcedger: Fixed — using `socket.end()` for graceful close.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,168 @@
...
+          venueId,
+          apikey: 'desktop_H2WRdpoSEh7iOWD2iCZD7msTKOs',
+        },
+      })
+    );
+  });
+
+  ws.on('message', (data: WebSocket.Data) => {
+    handleMessage(data.toString());
+  });
```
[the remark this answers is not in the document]
Knorcedger: Fixed — handling string, Buffer, and ArrayBuffer data types.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,168 @@
...
+    logger.info('WebSocket connected to backend');
+
+    // Register as printer server
+    ws!.send(
+      JSON.stringify({
+        type: 'printerServerRegister',
+        data: {
+          venueId,
+          apikey: 'desktop_H2WRdpoSEh7iOWD2iCZD7msTKOs',
+        },
```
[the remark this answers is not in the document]
Knorcedger: Fixed — API key now loaded from config via `nconf.get('API_KEY')` with fallback.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,168 @@
...
+  try {
+    const msg = JSON.parse(message);
+
+    switch (msg.type) {
+      case 'printRaw': {
+        const { jobId, printerIp, printerPort, data } = msg;
+        const buffer = Buffer.from(data, 'base64');
+        const port = printerPort ? parseInt(printerPort, 10) : 9100;
+
+        logger.info(`Received print job ${jobId} for ${printerIp}:${port} (${buffer.length} bytes)`);
```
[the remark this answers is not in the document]
Knorcedger: Fixed — validating required fields (jobId, printerIp, data) before processing.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,171 @@
...
+    ws!.send(
+      JSON.stringify({
+        type: 'printerServerRegister',
+        data: { venueId, apikey: getApiKey() },
+      })
+    );
+  });
+
+  ws.on('message', (data: WebSocket.Data) => {
+    const message = typeof data === 'string' ? data : Buffer.isBuffer(data) ? data.toString() : Buffer.from(data as ArrayBuffer).toString();
```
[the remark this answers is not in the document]
Knorcedger: Fixed — added `Array.isArray(data)` check with `Buffer.concat(data)` for chunked messages.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,171 @@
...
+  try {
+    const fs = require('fs');
+    if (fs.existsSync('./settings.json')) {
+      const settings = JSON.parse(fs.readFileSync('./settings.json', 'utf8'));
+      if (settings.venueId) return settings.venueId;
+      if (settings.modem?.venueId) return settings.modem.venueId;
+    }
+  } catch {}
+  return nconf.get('VENUE_ID') || '';
+}
```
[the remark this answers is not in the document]
Knorcedger: Fixed — venueId is now cached after first read. Uses the existing `getSettings()` module instead of re-reading settings.json.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,171 @@
...
+
+function getApiKey(): string {
+  return nconf.get('API_KEY') || 'desktop_H2WRdpoSEh7iOWD2iCZD7msTKOs';
```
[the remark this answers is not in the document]
Knorcedger: Fixed — using a shared `API_KEY` constant. It's the same key used in `modules/api.ts`. This is a desktop app key, not a secret — it's already in the codebase.

package.json
```diff
@@ -31,6 +31,7 @@
   "dependencies": {
     "@serialport/bindings-cpp": "^12.0.1",
     "@types/jszip": "^3.4.0",
+    "@types/ws": "^8.18.1",
     "adm-zip": "^0.5.15",
     "axios": "^1.11.0",
```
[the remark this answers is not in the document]
Knorcedger: Fixed — moved `@types/ws` to devDependencies.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,171 @@
...
+        sendToPrinter(printerIp, port, buffer)
+          .then(() => {
+            logger.info(`Print job ${jobId} sent successfully to ${printerIp}`);
+            sendResult(jobId, 'success');
+          })
+          .catch((err) => {
+            logger.error(`Print job ${jobId} failed for ${printerIp}:`, err.message);
+            sendResult(jobId, 'failed', err.message);
+          });
+        break;
```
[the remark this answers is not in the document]
Knorcedger: Acknowledged — tests will be added in a follow-up. The printer server runs on venue Windows PCs and integration testing requires a physical printer or mock TCP server.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,171 @@
...
+function sendResult(jobId: string, status: string, error?: string): void {
+  if (ws && ws.readyState === WebSocket.OPEN) {
+    ws.send(
+      JSON.stringify({
+        type: 'printRawResult',
+        data: { jobId, status, error, venueId: getVenueId() },
+      })
+    );
+  }
+}
```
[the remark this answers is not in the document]
Knorcedger: Fixed — `sendResult` now logs an error when the WebSocket is not open, so we have visibility into dropped results.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,185 @@
...
+    });
+    socket.setTimeout(SOCKET_TIMEOUT);
+    socket.on('error', (err) => {
+      socket.destroy();
+      reject(err);
+    });
+    socket.on('timeout', () => {
+      socket.destroy();
+      reject(new Error('Connection timeout'));
+    });
```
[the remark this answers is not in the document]
Knorcedger: Fixed — promise now resolves on `'close'` event instead of `write` callback, with a settle guard to prevent double resolve/reject from error+close race.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,185 @@
...
+    const settings = getSettings();
+    if (settings?.venueId) { cachedVenueId = settings.venueId; return cachedVenueId!; }
+    if (settings?.modem?.venueId) { cachedVenueId = settings.modem.venueId; return cachedVenueId!; }
+  } catch {}
+  cachedVenueId = nconf.get('VENUE_ID') || '';
+  return cachedVenueId!;
+}
+
+// Shared API key constant — same as used in modules/api.ts
+const API_KEY = 'desktop_H2WRdpoSEh7iOWD2iCZD7msTKOs';
```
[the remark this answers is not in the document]
Knorcedger: Fixed — API_KEY is now loaded from `nconf.get('API_KEY')` with hardcoded fallback. To fully centralize, would need to refactor `api.ts` too (it has the key inline in a curl string), which is out of scope for this PR.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,185 @@
...
+          logger.error('Invalid printRaw message: missing required fields');
+          if (jobId) sendResult(jobId, 'failed', 'Missing required fields');
+          return;
+        }
+
+        const buffer = Buffer.from(data, 'base64');
+        const port = printerPort ? parseInt(printerPort, 10) : 9100;
+
+        logger.info(`Received print job ${jobId} for ${printerIp}:${port} (${buffer.length} bytes)`);
+
```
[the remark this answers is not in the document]
Knorcedger: Fixed — using `Number.isFinite(parsed)` to validate the parsed port. Falls back to 9100 for NaN or non-numeric input.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,211 @@
...
+  if (cachedVenueId) return cachedVenueId;
+  try {
+    // Try existing settings module first
+    const { getSettings } = require('./settings');
+    const settings = getSettings();
+    if (settings?.venueId) { cachedVenueId = settings.venueId; return cachedVenueId!; }
+    if (settings?.modem?.venueId) { cachedVenueId = settings.modem.venueId; return cachedVenueId!; }
+  } catch {}
+  cachedVenueId = nconf.get('VENUE_ID') || '';
+  return cachedVenueId!;
```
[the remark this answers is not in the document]
Knorcedger: Fixed — using `!== null` instead of truthy check so empty string venueId is treated as cached.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,211 @@
...
+          .catch((err) => {
+            logger.error(`Print job ${jobId} failed for ${printerIp}:`, err.message);
+            sendResult(jobId, 'failed', err.message);
+          });
+        break;
+      }
+
+      case 'checkPrintersRequest': {
+        logger.info('Received printer check request');
+        const printersToCheck: { id: string; ip: string; port?: string }[] = msg.printers || [];
```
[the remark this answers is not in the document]
Knorcedger: Fixed — now reads from `msg.data.printers || msg.printers` to handle both nested and flat formats.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,211 @@
...
+
+        logger.info(`Received print job ${jobId} for ${printerIp}:${port} (${buffer.length} bytes)`);
+
+        sendToPrinter(printerIp, port, buffer)
+          .then(() => {
+            logger.info(`Print job ${jobId} sent successfully to ${printerIp}`);
+            sendResult(jobId, 'success');
+          })
+          .catch((err) => {
+            logger.error(`Print job ${jobId} failed for ${printerIp}:`, err.message);
```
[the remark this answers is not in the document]
Knorcedger: Fixed — logging full error object instead of just `err.message`.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,211 @@
...
+    handleMessage(message);
+  });
+
+  ws.on('close', () => {
+    logger.info('WebSocket connection closed');
+    scheduleReconnect();
+  });
+
+  ws.on('error', (err) => {
+    logger.error('WebSocket error:', err.message);
```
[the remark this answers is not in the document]
Knorcedger: Fixed — logging full error object for WS errors.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,552 @@
...
+// fs.writeFile to \\host\share reports success even when the physical
+// printer is offline or the share does not exist. Gate on the printer's
+// WMI WorkOffline state first (mirrors the legacy print path) so an
+// offline printer fails loudly instead of returning a false success.
+// Stricter than legacy: only an explicit `false` (share found AND not
+// offline) counts as online — empty output means "share not found".
+async function isWindowsSharedPrinterOnline(
+  shareName: string
+): Promise<boolean> {
+  const command = `powershell -NoProfile -Command "Get-WmiObject -Query \\"SELECT * FROM Win32_Printer WHERE ShareName = '${shareName}'\\" | Select-Object -ExpandProperty WorkOffline"`;
```
Knorcedger: **Duplication (cleanup).** This `isWindowsSharedPrinterOnline` is a near-verbatim copy of `isUsbPrinterOnline` in `src/modules/printer.ts:433-435` — same `powershell ... Win32_Printer WHERE ShareName = '${shareName}' ... WorkOffline` query. Prefer exporting/reusing the existing helper (or hoisting it to `common.ts`) so the WMI/quoting logic lives in one place.

Note both copies interpolate `shareName` straight into the PowerShell command unescaped. `shareName` derives from the backend-sent `printerPort`, so it's low-risk with a trusted BE, but a `'` in a share name would break the query and a hostile value could inject. Worth escaping the single-quote when you consolidate.

src/modules/api.ts
```diff
@@ -117,6 +117,37 @@ export const apiCall = async (query: string): Promise<any> => {
...
+// the GraphQL endpoint usually still works, and curl covers the proxy case.
+const reportWebSocketFailure = async (details: {
+  attempts: number;
+  category: string;
+  code: string;
+  message: string;
+  url: string;
+  venueId: string;
+}): Promise<void> => {
+  const message = `Problem: printer-server cannot open its WebSocket to ${details.url} for venue ${details.venueId} after ${details.attempts} attempts — ${details.category} (${details.code}): ${details.message}. Backend->printer dispatch is down for this venue.`;
```
Knorcedger, line 142: **Duplication (cleanup).** `reportWebSocketFailure` rebuilds the exact `addError` mutation shape as `reportFetchFailure` (lines 51-79): same `Problem:` prefix, same `escapeGraphqlString` calls, same `detailsJson`→`query` packing. Consider a shared `buildAddErrorMutation(message, url, details)` helper so the two don't drift apart.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,552 @@
...
+  const wsUrl = nconf.get('BACKEND_WS_URL');
+  if (wsUrl) return wsUrl;
+  return apiUrl
+    .replace('/graphql', '')
+    .replace('https://', 'wss://')
+    .replace('http://', 'ws://');
+}
+
+// Get registered venueId from in-memory settings object.
+function getVenueId(): string {
```
Knorcedger, line 120: **Minor.** `getVenueId()` / `getWsSecret()` do `require('./settings')` on every call, and `getVenueId()` runs on each inbound message and each result/check send. `settings.ts` doesn't import `wsClient`, so a top-level `import { getSettings } from './settings'` is safe here and avoids the repeated dynamic resolution. If the `require` is a deliberate circular-import dodge, a one-line comment would help.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,552 @@
...
+      ws.removeAllListeners();
+      ws.close();
+    } catch (err) {
+      logger.error('reconnectWebSocketClient: error closing old socket:', err);
+    }
+    ws = null;
+  } else {
+    logger.info('reconnectWebSocketClient: creds now available, connecting');
+  }
+  reconnectAttempts = 0;
```
Knorcedger, line 578: **Edge case.** `reconnectWebSocketClient()` resets `reconnectAttempts` and `authFailureLogged` but not the connect-failure trio (`consecutiveConnectFailures`, `connectionReportSent`, `lastWsError`) — those only reset via `stableTimer` after a socket stays open 5s. So if a creds-change reconnect still can't open the socket, a stale `lastWsError` from the prior episode can be attributed to the new close-before-open, and `connectionReportSent` staying `true` suppresses a fresh incident. Low severity (self-heals once a socket sticks), but consider resetting them here too.

Knorcedger (commented):
## Mike automated review

**Task link:** PR description links a Flarmio Partners task. No flarmio-partners tool is available in this environment, so I could verify the link is *present* but not that the PR is actually associated with that task in Partners — flagging as not-machine-verified.

**Verdict:** Changes not approved — 1 blocking security gap. (Also note: this PR is self-authored, so GitHub blocks me from submitting an approval regardless; manual approval required.)

### 🔴 Blocking — `wsSecret` still leaks via the unauthenticated `GET /` homepage

`src/homepage.ts:15` renders `${JSON.stringify(getSettings(), null, 2)}` directly into the HTML returned by `GET /` (`src/index.ts:104-106`), which is unauthenticated and reachable by anything on the venue LAN on port 7810.

This PR adds `wsSecret` as a per-venue credential and carefully strips it everywhere else — `getPublicSettings()` for `GET /settings`, the `POST /settings` response, and all the settings log dumps — but the homepage still serializes the **full** settings object, including `wsSecret`. That defeats the stated goal of commit `804b4af` ("conceal wsSecret from settings queries") and exposes the credential that authenticates this venue to the backend WS.

**Fix:** make `homepage()` use `getPublicSettings()` instead of `getSettings()`.

---

Non-blocking items left as inline comments (duplication / minor efficiency / reconnect state-reset edge case).

src/modules/wsClient.ts
Knorcedger: **Command injection + duplication.** `shareName` is derived from `printerPort`, which comes straight from the backend `printRaw` WS message (`deviceInterface.split('\\').pop()`), then interpolated unescaped into a `powershell -Command` string. A crafted `printerPort` (e.g. containing `'; <command>; #`) runs arbitrary PowerShell on the venue's Windows machine. The HTTP path's `shareName` comes from local config, but here it's remote-controlled, so the exposure is wider.

Also: this function is a byte-for-byte copy of `isUsbPrinterOnline` in `src/modules/printer.ts:433`. Reuse the existing helper (and fix the escaping there) rather than maintaining two copies of the same WMI command.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,552 @@
...
+  });
+}
+
+async function handleMessage(raw: string): Promise<void> {
+  try {
+    const msg = JSON.parse(raw);
+
+    switch (msg.type) {
+      case 'printRaw': {
+        const { jobId, printerIp, printerPort, data } = msg;
```
Knorcedger, line 248: **Message-shape inconsistency.** `printRaw` reads `jobId`/`printerIp`/`printerPort`/`data` from the top level of `msg`, but every other message in this file uses a `data` wrapper: `printerServerRegister` and `printRawResult`/`checkPrintersResponse` **send** `{ type, data: {...} }`, and `checkPrintersRequest` defensively reads `msg.data?.printers || msg.printers`. If the backend wraps `printRaw` payloads in `data` like the rest, every field is `undefined`, the job is silently dropped, and because `jobId` is undefined no `printRawResult` is ever sent back — the BE waits forever. Please confirm the exact backend envelope and read it consistently (mirror the `msg.data?.x || msg.x` fallback used just below).

src/modules/wsClient.ts
```diff
@@ -0,0 +1,552 @@
...
+  }
+
+  if (ws) {
+    logger.info(
+      credsChanged
+        ? 'reconnectWebSocketClient: creds changed, reconnecting'
+        : 'reconnectWebSocketClient: socket not live, reconnecting'
+    );
+    try {
+      ws.removeAllListeners();
```
Knorcedger, line 566: **Orphaned `stableTimer` on creds-changed reconnect.** `removeAllListeners()` here means the old socket's `close` handler never runs, so if that socket had already fired `open` and armed `stableTimer`, the timer is never cleared. ~5s later it fires against the new connection and resets `reconnectAttempts`/`consecutiveConnectFailures`/`lastWsError` mid-episode, corrupting the back-off accounting. Clear `stableTimer` (and `reconnectTimer`) explicitly before closing the old socket here.

Knorcedger (commented):
## Mike automated code review — changes needed

Reviewed `feat/ws-client` against `origin/main` via `/code-review` (high effort). One **blocking** issue plus a few worth addressing.

### 🔴 Blocking — `wsSecret` leaks on the unauthenticated `GET /` homepage
`src/homepage.ts:15` renders `JSON.stringify(getSettings(), null, 2)` directly into the HTML served at the root route (`src/index.ts:104-105`). This PR introduces `wsSecret` into the settings object and carefully strips it from `GET /settings` (`getPublicSettings`), the `POST /settings` response, and all logs — but the homepage still dumps the **full** settings, so the per-venue WS credential is exposed in plaintext to anyone who can reach port 7810. `homepage.ts` isn't in this diff, so I can't inline-comment it, but it must use `getPublicSettings()` (or otherwise omit `wsSecret`) before merge.

### Other findings
Left as inline comments below: command-injection / duplication in the Windows share check, a `printRaw` message-shape inconsistency, and a reconnect timer leak.

---
_Task link present in PR description (Flarmio Partners), but no flarmio-partners tool is available to confirm the PR is actually linked to that task — unverified. PR is self-authored, so GitHub approval is left to a human reviewer._

src/modules/wsClient.ts
Knorcedger: **Command injection: `shareName` is interpolated unescaped into a PowerShell command.**

`shareName` comes from `deviceInterface.split('\\').pop()`, where `deviceInterface` is the `printerPort` field of an inbound `printRaw` WS message. It is dropped verbatim into the `powershell -Command "...ShareName = '${shareName}'..."` string with no escaping. A device path whose share segment contains `'` or `"` (e.g. from a malformed/spoofed backend message or a venue printer configured with a crafted path) breaks out of the WQL string and can execute arbitrary PowerShell on the venue machine. Even though the WS link is authenticated, a field that builds a shell command should be validated/escaped (reject share names not matching e.g. `^[A-Za-z0-9 _.$-]+$`, or pass via a here-string/argument rather than string interpolation).

src/modules/wsClient.ts
Knorcedger, line 566: **`stableTimer` is leaked here, so a stale back-off reset can fire on the next socket.**

`removeAllListeners()` strips the `close` handler that is the only place clearing `stableTimer` (lines 430-432). If the socket being replaced opened less than `CONNECTION_STABLE_MS` ago, its `stableTimer` is still pending. `connect()` does not clear it either, and the next `open` overwrites the `stableTimer` variable — orphaning the old handle. The orphaned timer then fires and resets `reconnectAttempts`, `consecutiveConnectFailures`, `connectionReportSent`, etc. on the *new* connection mid-lifecycle, defeating the back-off it was meant to gate. Clear `stableTimer` here (and/or at the top of `connect()`) before closing/replacing the socket.

src/modules/wsClient.ts
Knorcedger, line 578: **Reconnect resets `reconnectAttempts`/`authFailureLogged` but not `connectionReportSent`/`consecutiveConnectFailures`.**

After a venue is firewalled long enough to raise the one-per-episode Slack incident (`connectionReportSent = true`, `consecutiveConnectFailures >= 3`), a settings sync calls `reconnectWebSocketClient()`, which clears `reconnectAttempts` and `authFailureLogged` for a clean reconnect but leaves `connectionReportSent`/`consecutiveConnectFailures` set. Only a *stable* (>5s) connection resets them via `stableTimer`. So if the new connection keeps failing without ever reaching the 5s stable window, the close handler sees `connectionReportSent` still `true` and never raises a second incident — the venue stays silently down. Reset these two here as well for episode parity.

src/resolvers/settings.ts
```diff
@@ -84,13 +89,22 @@ const settings = async (req: Request<{}, any, any>, res: Response<{}, any>) => {
       }
     }
 
-    logger.info('Settings updated:', newSettings);
+    // Never echo or log wsSecret: the FE already holds the value it pushed,
+    // and the response/logs must not expose the credential.
+    const { wsSecret: _wsSecret, ...safeSettings } = newSettings;
```
Knorcedger: **Cleanup: the `wsSecret`-stripping logic is now duplicated in three places.**

This PR adds `getPublicSettings()` in `settings.ts` to strip `wsSecret`, but the resolver re-implements the same `const { wsSecret, ...rest } = ...` inline here and at line 19, and `loadSettings()` does it again. A future second secret (or a rename of `wsSecret`) requires editing every site independently, and missing one re-leaks the credential. Consider a single shared `stripSecrets(settings)` helper used by all of them.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,552 @@
...
+        // A job needs an id, a payload, and at least one transport target:
+        // an IP for TCP printers, or a device path in printerPort for
+        // local shared/USB/serial printers.
+        if (!jobId || !data || (!printerIp && !printerPort)) {
+          logger.error('Invalid printRaw message: missing required fields');
+          if (jobId) sendResult(jobId, 'failed', 'Missing required fields');
+          return;
+        }
+
+        const buffer = Buffer.from(data, 'base64');
```
Knorcedger, line 266: **Minor: a non-string/un-decodable `data` leaves the job unacknowledged.**

The guard above only checks `!data`. `Buffer.from(data, 'base64')` then runs inside the outer `try` of `handleMessage`; if `data` is an object/array (malformed message) `Buffer.from` throws, the outer `catch` logs `"Error handling WS message"` but never calls `sendResult(jobId, 'failed')`. The backend is left waiting indefinitely for a result it will never receive. Consider validating `typeof data === 'string'` and emitting a failed `sendResult` on bad input.

src/modules/wsClient.ts
```diff
@@ -0,0 +1,552 @@
...
+}
+
+// A Windows shared-printer write is fire-and-forget at the spooler:
+// fs.writeFile to \\host\share reports success even when the physical
+// printer is offline or the share does not exist. Gate on the printer's
+// WMI WorkOffline state first (mirrors the legacy print path) so an
+// offline printer fails loudly instead of returning a false success.
+// Stricter than legacy: only an explicit `false` (share found AND not
+// offline) counts as online — empty output means "share not found".
+async function isWindowsSharedPrinterOnline(
```
Knorcedger: **Reuse: this is byte-for-byte identical to `isUsbPrinterOnline` in `printer.ts:433`.**

`printer.ts:433` already defines the exact same `powershell ... Win32_Printer WHERE ShareName = '${shareName}'` helper (including the same unescaped-`shareName` injection noted above). Rather than copy it into `wsClient.ts`, export and reuse the existing one — then the escaping fix lands in a single place and the two print paths can't diverge.

src/modules/wsClient.ts
Knorcedger, line 248: **Message-envelope inconsistency: `printRaw` is read from the top level, but every other message uses a `data` wrapper.**

Here `printRaw` destructures `jobId/printerIp/printerPort/data` from the top level of `msg`. But this same file *sends* results as `{ type, data: { jobId, ... } }` (`sendResult`, `checkPrintersResponse`), and `checkPrintersRequest` even reads `msg.data?.printers || msg.printers` to tolerate both shapes. If the backend dispatches `printRaw` wrapped in `data` (mirroring the result envelope it receives), every field here is `undefined`: the job is silently dropped and — because `jobId` is also `undefined` — no failure result is ever sent back, so the backend waits forever. Confirm the backend envelope and read `printRaw` the same way as the others.

src/modules/wsClient.ts
nina-dk: fixed in [15479a5](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/15479a50e63f661bfe6242e58930b48285b0c290).

src/modules/wsClient.ts
```diff
@@ -0,0 +1,576 @@
...
+            logger.error(`Print job ${jobId} failed for ${target}:`, err);
+            sendResult(jobId, 'failed', err?.message ?? String(err));
+          });
+        break;
+      }
+
+      case 'checkPrintersRequest': {
+        logger.info('Received printer check request');
+        const printersToCheck: { id: string; ip: string; port?: string }[] =
+          msg.data?.printers || msg.printers || [];
```
Knorcedger, line 307: Non-blocking, but this connectivity-check feature is currently a no-op against the real backend. The backend (`quickord-be@feat/print-system-refactor-v2`) sends `checkPrintersRequest` as `{ data: { venueId }, type: checkPrintersRequest }` (see `modules/printing/dispatcher.ts` and the typed contract in `modules/websockets/types.ts`) — there is **no** `printers` array in the message. So `msg.data?.printers || msg.printers || []` always resolves to `[]`, `results` is empty, and `checkPrintersResponse` reports zero printer statuses for the venue.

Suggest falling back to the locally-configured printers when the message omits a list, e.g. `const printersToCheck = msg.data?.printers || msg.printers || getSettings().printers || []` (mapping each printer to `{ id, ip, port }`). The printer-server already knows its own printers, which is presumably why the backend only sends `venueId`.

Knorcedger (commented):
## Re-review (post-latest-changes) — approval summary

**Verdict:** Approve (posted as a comment — GitHub blocks self-approval since I authored the PR).

I re-reviewed the latest commit `15479a5` against `origin/main`, cross-checking the WS envelope against the backend counterpart (`quickord-be@feat/print-system-refactor-v2`).

### Previous findings — all resolved ✅
- **wsSecret exposure (was blocking):** `/` homepage now serializes `getPublicSettings()`, `GET /settings` returns `getPublicSettings()`, and the `POST /settings` response/logs strip `wsSecret`. Verified.
- **PowerShell/WQL injection in shared-printer helper:** single-quote doubling (`replace(/'/g, "''")`) added.
- **`printRaw` envelope consistency:** verified against backend — backend sends flat `{ data: <base64>, jobId, printerIp, printerPort, type }`; client destructures flat. Matches. `printRawResult`, `printerServerRegister`, and `registered` (flat `venueId`) envelopes also match.
- **Reconnect timer cleanup:** `reconnectWebSocketClient()` now clears `stableTimer` before `removeAllListeners()` and resets the full incident state.

### One non-blocking finding (see inline comment)
`checkPrintersRequest` handler expects a printers array the backend never sends (backend's typed message is `{ data: { venueId } }`), so connectivity checks always return empty. Recommend falling back to the locally-configured printers. Secondary diagnostic feature; does not affect the core print path. Both repos are still on feature branches, so the contract can be aligned before merge.

src/modules/wsClient.ts
nina-dk: fixed in [15479a5](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/15479a50e63f661bfe6242e58930b48285b0c290).

src/modules/wsClient.ts
nina-dk: fixed in [15479a5](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/15479a50e63f661bfe6242e58930b48285b0c290).

src/modules/wsClient.ts
nina-dk: fixed in [15479a5](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/15479a50e63f661bfe6242e58930b48285b0c290).

src/modules/wsClient.ts
nina-dk: fixed in [15479a5](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/15479a50e63f661bfe6242e58930b48285b0c290).

src/modules/wsClient.ts
nina-dk: [c969896](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/c969896aaa7df7068576f28e0cf7b791ca9c8ec8)

src/modules/wsClient.ts
nina-dk: [c969896](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/c969896aaa7df7068576f28e0cf7b791ca9c8ec8)

src/resolvers/settings.ts
nina-dk: fixed in [15479a5](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/15479a50e63f661bfe6242e58930b48285b0c290).

src/modules/wsClient.ts
nina-dk: fixed in [15479a5](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/15479a50e63f661bfe6242e58930b48285b0c290).

src/modules/wsClient.ts
nina-dk: fixed in [15479a5](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/15479a50e63f661bfe6242e58930b48285b0c290).

src/modules/api.ts
nina-dk: fixed in [15479a5](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/15479a50e63f661bfe6242e58930b48285b0c290).

src/modules/wsClient.ts
nina-dk: fixed in [15479a5](https://github.com/Knorcedger/quickord-printer-server/pull/74/commits/15479a50e63f661bfe6242e58930b48285b0c290).

PR #98, opened 2026-05-26
feat: print new transfer product state

nina-dk:


Knorcedger (commented):
Found one blocking issue:

- `lastEditStatus: 'TRANSFERRED'` products can still be filtered out before this new print branch runs. In `printOrder`, edit orders first reduce `productsToPrint` to products whose `updateStatus` includes `NEW` or `UPDATED`. If a transferred product keeps `updateStatus` as `INITIAL`/`UNCHANGED` and only sets `lastEditStatus: 'TRANSFERRED'`, it is skipped and an edit containing only transferred products exits as “No NEW or UPDATED products to print”. The edit filter should also keep transferred products (and the skip reason may need updating).

PR #99, opened 2026-05-27
fix: make option & choice titles optional and nullable

nina-dk:


Knorcedger (changes requested):
Found one issue:

- [P2] Accept nullable choice titles in the validator — `src/resolvers/printOrders.ts:88-92`

When a null or missing title appears on an option choice (`options[].choices[].content[]`) rather than on the option group's own `content`, this endpoint still rejects the order in `Orders.parse` because the nested choice schema remains `title: z.string()`. The updated `OptionChoice` type in `common.ts` now permits nullable/missing choice titles and `getTitle` handles them, so the runtime validator should be relaxed for the same field or these orders will still fail before printing.

Reviewed with Codex against `origin/main`; no tests/build run as part of review.

Knorcedger (approved):
Re-reviewed after the latest commit. The validator now accepts nullable/missing titles for both option-group content and choice content, matching the widened local interfaces, and `getTitle` already falls back safely.

No remaining issues found. Reviewed with Codex against `origin/main`; no tests/build run as part of review.

PR #100, opened 2026-05-28
feat: bold table number along with order type

nina-dk:


Knorcedger (approved):
Reviewed with Codex and a manual pass over the printer formatting changes. I did not find any blocking issues.

Notes:
- Existing table/waiter formatting path is preserved when `BOLD_ORDER_TYPE` is off.
- The new enlarged table-number path restores text size before continuing, so later order type/product output should keep the configured size.
- I attempted `npm run typecheck`, but dependencies are not installed in the review checkout (`tsc: not found`), so I could not complete local typecheck here.

PR #101, opened 2026-06-09
feat: support full order prints

nina-dk:
[task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&task=6a21529936651eefb8851590)


Knorcedger (changes requested):
Mike review result: **Request changes**.

The core full-order rendering approach is good, but I found blocking/correctness issues that should be fixed or explicitly resolved before merge:

1. `config.json` changes `QUICKORD_API_URL` from production to `http://localhost:4000/graphql`. This looks accidental and would break shipped venue machines. Please revert to `https://api.quickord.com/graphql`.

2. FULL-ORDER prints are skipped by default. `documentsToPrint` is defaulted by schema and the default list does not include `FULL-ORDER`, so default-configured printers will skip full-order jobs. If this is intentional opt-in, confirm the FE/settings flow always adds `FULL-ORDER`; otherwise bypass or adjust this gate for full-order prints.

3. Customizations are not guaranteed. Option details are only printed when `OPTION-DETAILS` is in `documentsToPrint`, but the task requires customizations on the full-order ticket. FULL-ORDER mode should force option detail rendering or otherwise guarantee customizations are included.

Non-blocking: `printFullOrders` duplicates most of `printOrders`; consider extracting a shared handler.

Knorcedger (changes requested):
Mike rereview result: **Request changes**.

The previous blockers were mostly fixed:
- `config.json` is back to `https://api.quickord.com/graphql` ✅
- Full-order mode now forces `OPTION-DETAILS`, so customizations are guaranteed ✅
- Eligible printers with `FULL-ORDER` in `documentsToPrint` should now pass the document gate ✅

One new blocker was introduced in the fix commit:

`src/index.ts` now registers `/print-full-order` with `dedup`:

```ts
app.route('/print-full-order').post(dedup, printFullOrders);
```

The dedup middleware skips identical requests to the same endpoint within 30 seconds. Full-order printing is an explicit on-demand/reprint action from the UI, so if the user presses the button twice for the same order, the second print can be silently skipped as `duplicate request`. The original implementation correctly avoided dedup here with the comment that FULL-ORDER must reprint.

Please restore `/print-full-order` without `dedup`:

```ts
app.route('/print-full-order').post(printFullOrders);
```

Non-blocking: `src/modules/settings.ts` also removes `RATEUS` from the default `documentsToPrint`; I did not treat it as a blocker for this task, but please confirm it is intentional because it is unrelated to full-order printing.

Knorcedger (changes requested):
## Re-review (Mike) — `feat: support full order prints`

Re-checked the three previously-raised blockers. **All three are resolved:**

1. **`config.json` API URL** — ✅ Reverted to `https://api.quickord.com/graphql` on this branch; no localhost leftover.
2. **FULL-ORDER opt-in gate** — ✅ Confirmed intentional. The `requiredDocument = 'FULL-ORDER'` check (`printer.ts:3034`) is the correct opt-in: only printers the FE configures with `FULL-ORDER` in `documentsToPrint` are eligible, and the override deliberately injects only `OPTION-DETAILS` (not `FULL-ORDER`). Default printers are intentionally not eligible. Working as designed.
3. **Customizations guaranteed** — ✅ `OPTION-DETAILS` is injected into the full-order settings override, so the gate at `printer.ts:3318` is always satisfied on full-order tickets regardless of the printer's configured option flags.

Also verified: `documentsToPrint` schema is `z.array(z.string())` so `FULL-ORDER`/`OPTION-DETAILS` validate fine; `dedup` keys on `req.path` so the new route can't collide with `/print-orders`; `RATEUS` is unreferenced anywhere so dropping it from the default is safe cleanup; resolver error-handling/status parity with `printOrders` is intact.

### One new correctness gap (please address)

**`src/modules/printer.ts` — edit-mode display logic isn't neutralized in FULL-ORDER mode.**

The PR correctly guards the isEdit *filtering* with `!isFull` (`printer.ts:3031`), but the in-loop *display* logic still keys off the raw `isEdit` flag:

- `printer.ts:3309` — `if (isEdit) { ... }` prints NEW / UPDATED / MOVED / "quantity reduced" status markers.
- `printer.ts:3353-3367` — for an UPDATED product with `quantityChanged`, it overwrites the line with the **delta** (`is - was`): `productLine = \`${diff}x ...\`` and `printQuantity = diff`. The per-line total at `printer.ts:3406` (`(total + choicesTotal) * Math.abs(printQuantity)`) is then computed from that delta.

**Failure scenario:** a user edits an order (so the `/print-full-order` payload carries `isEdit: true`) and prints the full order. A product that went 3→5 prints as `2x <title>` priced for 2 units, and edit markers appear — even though the full-order ticket is meant to show the complete order with every item's true quantity, price, ΦΠΑ, and total. The grand total no longer matches the displayed line items.

This is the same intent the PR already applied to the filter — the display blocks just weren't covered. Suggested minimal fix, consistent with the existing pattern:

```ts
const showEditMarkers = !isFull && isEdit;
// then use showEditMarkers at printer.ts:3309 and in the quantityChanged condition at :3353
```

Everything else looks good. Nice job tightening 1–3. Once the edit-display blocks are guarded for full-order mode, this is good to merge.

Knorcedger (approved):
Mike rereview update: **Approved**.

Antonina clarified that `dedup` on `/print-full-order` is intentional to prevent duplicate full-order prints when multiple tabs trigger the same request. Given that product/operational context, I no longer consider the 30-second duplicate-skip behavior a blocker.

Previous findings status:
- `config.json` is back to the production Quickord API URL ✅
- Full-order mode now forces `OPTION-DETAILS`, so customizations are included ✅
- Eligible printers with `FULL-ORDER` in `documentsToPrint` pass the document gate ✅
- `dedup` on `/print-full-order` is intentional duplicate-print protection ✅

Non-blocking note: `RATEUS` was removed from the default `documentsToPrint`; please confirm that unrelated default change is intentional, but I am not blocking this PR on it.

PR #102, opened 2026-06-09
feat: hide per-product VAT on order when vat analysis is off

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/quickord/tasks?savedView=6a0d8b33cdba6716a23271b8&task=6a27fef1cb97d9bc4a5e13db)

### DIFF
--- a/src/modules/printer.ts
+++ b/src/modules/printer.ts
@@ -3403,12 +3403,17 @@ export const printOrder = async (
             (product.vat && settings.priceOnOrder === undefined) ||
             settings.priceOnOrder === true
           ) {
-            printer.println(
-              tr(
-                `${translations.printOrder.vat[lang]}: ${product.vat}%`,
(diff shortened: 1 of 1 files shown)


Knorcedger (approved):

I verified the linked Flarmio Partners task and reviewed the diff with Claude Code `/code-review`.

Summary:
- The change correctly gates the per-product `VAT: x%` print line behind the existing `vatAnalysis` setting.
- It matches the sibling guards already used for the VAT breakdown table and net-without-VAT total, so all related VAT analysis output now hides consistently when disabled.
- Behavior is unchanged when `vatAnalysis` is `true` or `undefined`.
- `vatBreakdown` accumulation remains outside the display guard, which keeps totals/data flow intact.

No findings. I did not run tests/build/lint, per review-only constraints.

PR #103, opened 2026-06-09
feat: pass issuerText to parkingTickets for custom headers

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/quickord/tasks?savedView=6a0d8b33cdba6716a23271b8&task=6a1eda6a1426408ace6f8cca)

### DIFF
--- a/src/modules/printer.ts
+++ b/src/modules/printer.ts
@@ -838,6 +838,7 @@ export const parkingTicket = async (
       req.body.phone,
       req.body.date,
       req.body.entryTime,
+      req.body.issuerText || '',
       //  req.body.operatingHours,
       req.body.lang || 'el'
     );
@@ -875,6 +876,7 @@ const printParkingTicket = async (
   phone: string,
   date: string,
   entryTime: string,
+  issuerText: string = '',
   // operatingHours: string,
   lang: SupportedLanguages = 'el'
 ) => {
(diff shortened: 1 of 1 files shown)


src/modules/printer.ts
```diff
@@ -917,11 +919,15 @@ const printParkingTicket = async (
         printer.println('PARKING TICKET');
         drawLine2(printer);
         printer.bold(false);
-        printer.bold(true);
-        printer.println(venueName);
-        printer.bold(false);
-        printer.println(address);
-        printer.println(phone);
+        if (issuerText) {
+          await readMarkdown(issuerText, printer, 'center', settings, true);
```
Knorcedger, line 923: Non-blocking nit: `readMarkdown` terminates its final text segment with `printer.print()` (no newline) when `issuerText` does not end in `\n`, whereas the old `else` branch ended with `printer.println(phone)`. The very next call here is `drawLine2(printer)`, so a single-line `issuerText` (e.g. `"My Parking Co"`) would get the divider glued onto the same physical line: `MY PARKING CO------...`. The else branch doesn't have this problem because it ends on a `println`.

This is the same latent asymmetry that already exists in `venueData`, so it's consistent with current code — but worth either documenting that `issuerText` must end in a newline, or adding a `printer.newLine()` after the `readMarkdown` call to make the layout robust to single-line headers.

Knorcedger (approved):
Approving. Clean, minimal change that correctly mirrors the existing `issuerText` pattern from `venueData` (common.ts:1010): the new `printParkingTicket` param is appended before `lang`, and the single call site is updated in the right order. Signature, defaulting (`= ''`), and the `req.body.issuerText || ''` guard all line up. `printParkingTicket` has only this one caller, and `readMarkdown(issuerText, printer, 'center', settings, true)` matches the helper signature.

One non-blocking nit left inline about line termination.

PR #104, opened 2026-06-10
feat: remove unused vulnerable axios version

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&task=6a296704b1d01dc9db9a7d67)


Knorcedger (approved):
Approve ✅

Verified `axios` is unused anywhere in `src/`, `scripts/`, and the root config. The only real code reference was the `package.json` declaration itself. Outbound HTTP in the app is done via `curl`/`exec` (`autoupdate.ts`, `api.ts`) and global `fetch`, none of which touch axios, so removal is safe at runtime and for the auto-update flow.

Lockfile is consistent: axios's private subtree (`follow-redirects`, `form-data`, `combined-stream`, `asynckit`, `delayed-stream`, `es-set-tostringtag`, `has-tostringtag`, `proxy-from-env`) is removed with no remaining references in the new lockfile, and `mime-types` is correctly retained for `express`. No other package versions changed; version file bumped correctly.

One non-blocking nit: the lockfile also drops `"peer": true` from 7 unrelated dev-dep nodes (babel/jest/typescript/etc.), which looks like npm-version regeneration churn rather than an intended change. Harmless, but consider regenerating with the team's standard npm to keep future diffs clean.

Note: this was a read-only review; I did not run `npm ci`, build, or typecheck. Recommend a quick `npm ci && npx tsc --noEmit` as a final local gate before merge.

PR #105, opened 2026-06-12
fix: get edit status from lastEditStatus flag first; add scripts to f…

nina-dk:
…ind greek charset/codepage combo on new printers

[task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&task=6a2c010970ceb4409b67a9c7)

### DIFF
--- a/scripts/greek-scan.js
+++ b/scripts/greek-scan.js
@@ -1,0 +1,93 @@
+// Greek codepage scanner for Aclas PP7X (and similar ESC/POS printers)
+// Node.js version - no PowerShell needed. Sends raw bytes to the printer over TCP 9100.
+//
+// Usage:
+//   node scripts/greek-scan.js                 (uses default IP below)
+//   node scripts/greek-scan.js 192.168.88.5    (pass a different IP)
+//   node scripts/greek-scan.js 192.168.88.5 255   (also raise MaxN if 63 finds nothing)
+//
+// Read the printed receipt: find the line where the Greek sample (ΑΒΓΔΕ αβγδε)
+// prints CORRECTLY. The number on that line is the value to use.
+//   - "t=NN" lines  -> use that NN as the `codePage` (ESC t n) value
+//   - the section header tells you which `characterSet` (encoding) matches
+
+import net from 'net';
+import iconv from 'iconv-lite'; // ships with node-thermal-printer
+
+const IP = process.argv[2] || '192.168.1.130';
+const PORT = 9100;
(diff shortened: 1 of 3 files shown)


Knorcedger (changes requested):
I found one blocking issue before this can be approved.

Blocking:
- `src/modules/printer.ts:2931`: `getEditStatus` says it prefers `lastEditStatus` and falls back to `updateStatus`, but it only returns actionable statuses (`NEW`, `UPDATED`, `TRANSFERRED`). If `lastEditStatus` is present as `UNCHANGED` or `INITIAL`, the helper skips it and falls back to `updateStatus`. That can still print/label an item as `NEW` or `UPDATED` based on stale `updateStatus`, which is exactly the class of bug this PR is meant to avoid. The effective status should likely be `product.lastEditStatus ?? product.updateStatus`, then the callers can decide whether to print/filter for `NEW`/`UPDATED`/`TRANSFERRED`.

Non-blocking notes:
- The new scanner scripts are useful, but `scripts/greek-scan.js` imports `iconv-lite` without the repo declaring it directly in `package.json`; it may work because of a transitive dependency, but a direct dependency or a note about using the app install context would make this less fragile.
- The JS and PowerShell scripts use different default printer IPs (`192.168.1.130` vs `192.168.88.5`). If intentional, fine; otherwise it is worth aligning them to reduce operator mistakes.

Suggested focused check:
- Add or manually verify an edited order product where `lastEditStatus: 'UNCHANGED'` and `updateStatus: 'NEW' | 'UPDATED'`. It should not pass the edit-print filter or show NEW/UPDATED labels if `lastEditStatus` is the authoritative effective flag.

Knorcedger (approved):
Follow-up after clarification: fallback to `updateStatus` is intentional for compatibility, and the payload lifecycle should not produce a product where `lastEditStatus` is `UNCHANGED` while `updateStatus` carries a newer actionable change.

Approved. The implementation matches the task: edit-print filtering, labels, and quantity-diff logic now use the effective edit status with `lastEditStatus` preferred and `updateStatus` retained as fallback.

Non-blocking notes from the first pass still stand, but they should not block merge:
- `scripts/greek-scan.js` relies on `iconv-lite`, which is currently available transitively through `node-thermal-printer`; declaring it directly would make the helper script less fragile.
- The JS and PowerShell scanner scripts use different default IPs. Fine if intentional, otherwise worth aligning later.

PR #106, opened 2026-06-15
feat: new setting to print menu categories on order & make categories…

nina-dk:
… bold

[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&task=6a2a7812f01cd33a7b5505f0)


Knorcedger (changes requested):
Requesting changes on one behavior issue before merge.

**Finding (medium): category header can show a category this printer is not configured to print**

In `src/modules/printer.ts` around the new grouping loop (`order.categoriesOrder.forEach(...)`, lines ~3544-3556), products are grouped under the first matching category from the full menu order, but the loop does not re-check the printer's `categoriesToPrint` allow-list.

Example: a product belongs to `[DRINKS, PROMO]`, and this kitchen printer is configured with `categoriesToPrint = [DRINKS]`. The product passes the earlier product filter because it has `DRINKS`. But if `PROMO` appears before `DRINKS` in `categoriesOrder`, it will print under a `PROMO` header, even though this station is explicitly not configured for that category.

Suggested fix: when `categoriesToPrint` is set, restrict category headers/grouping to allowed category ids, so the item prints under its first allowed category, or falls into `OTHER` only when none of its categories are allowed/in the order list.

Notes:
- I did not run builds/tests, per review policy.
- The rest of the implementation looks sound: flat fallback is guarded, `categoriesOrder` is optional, legacy settings default safely, and the render refactor preserves the per-product print behavior.
- Suggested regression test: multi-category product `[DRINKS, PROMO]`, printer `categoriesToPrint=[DRINKS]`, `categoriesOrder=[PROMO, DRINKS]`, expect it not to print under `PROMO`.

Knorcedger (approved):
Re-reviewed after `c5c4d0d`.

The previous blocker is resolved: grouping now skips category headers that are not allowed by this printer's `categoriesToPrint`, so multi-category products won't print under an excluded station/category header.

I also checked production data for the related question: there *are* products with more than one category, but it is rare (9 products across 6 menus, max 2 categories per product). So the edge case can happen in practice, and this guard is worth keeping.

No further blocking findings from this pass. I did not run builds/tests per review policy.

PR #108, opened 2026-06-24
fix: ping socket client-side to detect zombie sockets on PS side

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&task=6a3b88ddaa0536ee99b56090)


Knorcedger (approved):

I ran `/code-review` read-only against the PR and cross-checked the socket heartbeat behavior against the backend pattern. No blocking issues found.

Non-blocking notes:
- `src/modules/wsClient.ts:49`: comment says backend heartbeat is 25s, but backend appears to use 50s. Behavior is still fine because 20s < backend interval, but the comment should be corrected to avoid future confusion.
- `src/modules/wsClient.ts:454-457`: the missed-pong `terminate()` branch relies on the async `close` handler to clear the interval. Benign, but clearing `heartbeatTimer` before `sock.terminate()` would match the backend's more defensive pattern.

Good to merge from my side.

PR #109, opened 2026-06-24
feat: correct labels for APY + TPY printing

nina-dk:
[task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&task=6a354b6bb2c4dda467a1e2a9)

### DIFF
--- a/src/modules/printer.ts
+++ b/src/modules/printer.ts
@@ -1941,6 +1941,9 @@ const printPaymentReceipt = async (
         // Determine invoice type label based on AADE code
         let invoiceTypeLabel: string;
         switch (aadeInvoice?.header?.code) {
+          case '11.2':
+            invoiceTypeLabel = translations.printOrder.serviceReceipt[lang];
+            break;
           case '11.4':
             invoiceTypeLabel = translations.printOrder.retailCreditNote[lang];
             break;
@@ -2019,7 +2022,9 @@ const printPaymentReceipt = async (
         printer.newLine();
         printer.println(
           tr(
-            `${translations.printOrder.recieptEnd[lang]}`,
+            aadeInvoice?.header?.code === '11.2'
+              ? `${translations.printOrder.serviceReceiptEnd[lang]}`
+              : `${translations.printOrder.recieptEnd[lang]}`,
             settings.transliterate
           )
         );
@@ -2428,7 +2433,9 @@ const printMyPelatesReceipt = async (
         printer.newLine();
         printer.println(
           tr(
-            `${translations.printOrder.recieptEnd[lang]}`,
+            aadeInvoice?.header?.code === '11.2'
+              ? `${translations.printOrder.serviceReceiptEnd[lang]}`
+              : `${translations.printOrder.recieptEnd[lang]}`,
             settings.transliterate
           )
         );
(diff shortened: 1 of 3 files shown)


src/modules/printer.ts
```diff
@@ -2428,7 +2433,9 @@ const printMyPelatesReceipt = async (
         printer.newLine();
         printer.println(
           tr(
-            `${translations.printOrder.recieptEnd[lang]}`,
+            aadeInvoice?.header?.code === '11.2'
```
Knorcedger, line 2441: **Header/footer label mismatch in `printMyPelatesReceipt`.**

This footer is now conditioned on code `11.2` (prints "END OF SERVICE RECEIPT"), but the document **title** in this same function (`printer.ts:2358-2360`) is still hardcoded to `translations.printOrder.reciept` ("ΑΠΟΔΕΙΞΗ ΛΙΑΝΙΚΗΣ ΠΩΛΗΣΗΣ" / RETAIL RECEIPT) with no switch on `header.code`.

Result: a `11.2` service receipt printed via this path shows **RETAIL RECEIPT** at the top but **END OF SERVICE RECEIPT** at the bottom.

`printPaymentReceipt` got this right — it switches on `header.code` for **both** the title (`printer.ts:1943-1953`) and the footer. The MyPelates variant only received the footer change. Consider switching the title here on `11.2 → serviceReceipt` for consistency.

src/translations.ts
```diff
@@ -376,6 +376,22 @@ export default {
...
+    },
+    serviceReceiptEnd: {
+      el: 'ΑΠΟΔΕΙΞΗ ΠΑΡΟΧΗΣ ΥΠΗΡΕΣΙΩΝ ΛΗΞΗ',
+      en: 'END OF SERVICE RECEIPT',
+    },
+    serviceInvoice: {
+      el: 'ΤΙΜΟΛΟΓΙΟ ΠΑΡΟΧΗΣ ΥΠΗΡΕΣΙΩΝ',
+      en: 'SERVICE INVOICE',
+    },
+    serviceInvoiceEnd: {
```
Knorcedger, line 391: **`serviceInvoiceEnd` is dead code — added but never referenced.**

A grep across `src/` finds only this definition; nothing reads `translations.printOrder.serviceInvoiceEnd`. The invoice print paths (`getInvoiceTypeLabel` usages at `printer.ts:2180` and `2571`) print a title but no "end of document" footer, so there is no site that selects a service-vs-plain invoice end label. This mirrors the pre-existing unused `invoiceEnd` key just below.

Not blocking — either wire it into an invoice footer or drop it to avoid leaving unused translations.

Knorcedger (approved):
Reviewed via /code-review (high effort).

The core mapping is correct: AADE codes 2.1/2.2/2.3 → service invoice, 11.2 → service receipt, matching ΑΠΥ/ΤΠΥ semantics, and the new EL/EN labels follow the existing "... ΛΗΞΗ" footer convention.

Two non-blocking notes left inline:
1. `printMyPelatesReceipt` footer now prints "END OF SERVICE RECEIPT" for code 11.2, but its title is still hardcoded to the retail-receipt label (unlike `printPaymentReceipt`, which switches both). Header/footer can disagree for an 11.2 doc on this path.
2. `serviceInvoiceEnd` translation is added but never referenced (no invoice footer path uses it).

Neither blocks merge. Approving.

src/modules/printer.ts
nina-dk: fixed in [cd1a38f](https://github.com/Knorcedger/quickord-printer-server/pull/109/commits/cd1a38fee06a4e7ba5e82376355a11fc5f53d05c)

PR #110, opened 2026-06-24
feat: add stats printing as ne setting in documentsToPrint

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&task=6a32950db7f0cdd0253fc705)

### DIFF
--- a/src/modules/printer.ts
+++ b/src/modules/printer.ts
@@ -647,7 +647,8 @@ const printTextFunc = async (
   text: string,
   alignment: 'left' | 'center' | 'right',
   copies: number = 1,
-  lang: SupportedLanguages = 'el'
+  lang: SupportedLanguages = 'el',
+  requiredDocument: string = 'TEXT'
 ) => {
   let successCount = 0;
   const errors: Array<{ printerIdentifier: string; error: unknown }> = [];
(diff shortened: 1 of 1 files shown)


Knorcedger (commented):
Mike code review (read-only, against `main`). I found two minor issues worth addressing/confirming:

1. `STATS` is not present in the default `documentsToPrint` list. With the current implementation, a caller using `documentType: 'STATS'` will be skipped on printers that still have the default documents config; if no printer has been explicitly configured with `STATS`, `/print-text` fails with `Failed to print text to any printer`. Please confirm the rollout adds/exposes `STATS` in printer settings, or update the defaults/schema as needed.

2. `req.body.documentType` is passed through without validation, unlike `alignment`. A typo such as `STAT` will make every printer skip and return the generic print failure instead of a clear 400/validation error. Consider validating against supported document types or returning a clearer error.

I also verified the PR description contains a flarmio-partners task link.

Knorcedger (approved):
Reviewed the stats printing setting changes. I did not find code issues.

Task link present: https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&task=6a32950db7f0cdd0253fc705

Note: no local build/test/lint was run as part of this read-only review.

PR #112, opened 2026-06-30
feat: print order QR to assign delivery person

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&task=6a4224d7436ce54500b2c539)


src/modules/printer.ts
```diff
@@ -3753,6 +3753,37 @@ export const printOrder = async (
             )
           );
         }
+
+        // Delivery-assignment QR: a delivery person scans it to open the order
+        // and claim it. Present only for delivery orders when the venue setting
+        // is on (gated by the frontend before the order reaches here).
+        if (order.assignmentQrUrl) {
```
Knorcedger, line 3760: The only in-code gate here is `if (order.assignmentQrUrl)` — there is no check on printer role/type or copy index. This block lives inside the per-printer × per-copy loop, so the "SCAN TO CLAIM THIS ORDER" QR will print on **every** printer that prints this order ticket (kitchen/prep/bar stations included), and on every copy.

The comment says gating is "done by the frontend," but the frontend sends a single order payload that the server fans out to all matching printers — it cannot select *which* of this server's printers gets the QR. So if the claim QR is meant to appear only on a dispatch/cashier copy, this design will over-print it onto kitchen tickets too.

Is printing the claim QR on every station ticket intended? If not, the gate likely needs to live server-side (e.g. only on the printer that carries the full order, or a dedicated dispatch role), since the frontend cannot express per-printer targeting through this path. Either way, please confirm — happy to approve once this is intentional/documented.

Knorcedger (commented):
Automated review (Mike) — head `55cbac7`.

The QR feature is clean and well-implemented: the notice goes through `tr(..., settings.transliterate)` per the repo convention, the QR payload (a URL) is correctly left un-transliterated, `printQR` options match the existing AADE QR usage, alignment is reset on every path (incl. the catch), and the `z.string().optional()` guard plus `if (order.assignmentQrUrl)` correctly skip empty values. No crashes or type issues.

One concern worth confirming before merge (inline). Not approving pending the author's call on it.

src/modules/printer.ts
Knorcedger: Re-checked on `e940409` (the *fix: fix order QR layout* commit). That commit only reorders the QR block above the POWERED BY / AADE-notice section and centers the POWERED BY line — it does **not** add any per-printer or copy-index gate. So this point is still open: the claim QR continues to print on every matching station printer (kitchen/bar included) and on every copy. Could you confirm whether printing it on all station tickets is intended? Happy to approve once it's confirmed intentional or gated server-side.

src/modules/printer.ts
```diff
@@ -3730,8 +3730,39 @@ export const printOrder = async (
           );
         }
 
+        // Delivery-assignment QR: a delivery person scans it to open the order
+        // and claim it. Present only for delivery orders when the venue setting
+        // is on (gated by the frontend before the order reaches here).
+        if (order.assignmentQrUrl) {
+          printer.newLine();
+          printer.alignCenter();
+          printer.println(
```
Knorcedger, line 3739: Minor (non-blocking): the notice `println(... assignQrNotice ...)` is emitted here, *before* the `try`. `printQR` builds the QR bytes synchronously and can throw on bad input (e.g. a long signed delivery deep-link that overflows the symbol capacity at `correction: 'Q'`, `model: 4`). When it does, the `catch` logs and continues, so the receipt prints "SCAN TO CLAIM THIS ORDER" with no QR beneath it — an orphaned, un-scannable instruction. Consider moving the notice `println` inside the `try` after `printQR` (or only printing it on success) so the notice and QR succeed or fail together. Also note the "Failed to print" log can only fire for synchronous encode errors — actual transmission failures happen later in `executePrinter`, outside this `try`.

Knorcedger (approved):
Mike review approved.

I verified the Flarmio Partners task link, ran Claude Code /code-review, and found no blocking issues. The QR printing follows the existing AADE QR pattern and the added schema/translation changes match repo conventions.

PR #115, opened 2026-07-02
fix: improve network scan & printers status response times

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=69fdb53afa09045de57bf516&project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&task=6a452079c6ca86f2719be09b)


src/modules/network.ts
```diff
@@ -55,38 +86,21 @@ export default async function scanNetworkForConnections(): Promise<
...
+  const portResults = await runPool(targets, SCAN_CONCURRENCY, ({ ip, port }) =>
+    checkPort(ip, port)
+  );
 
-  // Step 3: Filter and collect open ports
+  const printers: { ip: string; port: number }[] = [];
   for (const result of portResults) {
     if (result.open) {
-      printers.push({ id: result.ip, ip: result.ip, port: result.port });
+      printers.push({ ip: result.ip, port: result.port });
```
Knorcedger, line 103: The scan results used to carry `id: result.ip` over the wire (both the `/network` HTTP response and the `scanNetworkResponse` WS payload), even though the declared return type only promised `{ ip, port }`. Dropping `id` is fine on the type level, but can any frontend/backend consumer still key scanned devices on `.id`? Since `id` was always equal to `ip`, worth a quick confirm that the FE list rendering / dedup doesn't rely on it before merging.

Knorcedger (approved):

Task link verified: https://partners.flarmio.com/admin/dev/tasks/6a452079c6ca86f2719be09b (`quickord`, P1 bug).

I reviewed the diff with Claude Code `/code-review` and checked the PR manually. No blocking correctness or regression issues found.

Summary:
- Direct TCP scan with bounded concurrency is a good replacement for the Windows `ping.exe` fan-out and should keep discovery well below the backend/Heroku timeout budget.
- Parallel `checkPrinters` keeps response shape/order while avoiding sequential ~19s-per-printer worst cases.
- `version` lookup now matches the packaged/service runtime path, with the old `__dirname` path retained as fallback.

Non-blocking follow-ups:
- `ping` and `@types/ping` are now unused and can be removed from `package.json` in a cleanup.
- Existing debounce keying on `settings.id || ''` can collide for id-less printers; not introduced by this PR, but worth a follow-up if USB/id-less configs are common.

CI note: GitHub reports no checks for this branch.

src/modules/network.ts
nina-dk: Confirmed safe — no consumer keys on `.id`, and it never reached the FE anyway:

- **BE strips it before the wire:** `requestNetworkScan` is typed `{ ip, port }[]` (dispatcher.ts) and the GraphQL `LanConnectionType` exposes only `ip`/`port` (scanPrinterNetwork.ts), so `id` was dropped at serialization even when the PS sent it.
- **FE never reads it:** `ScanPrinterNetwork.gql` selects only `ip`/`port`; the one consumer (`printer-settings.tsx` `refreshConnections`) dedups by `p.ip` and generates its own `makePrinterId()` for the row — it never touches a scan `.id`.
- **Android** does not use the scan feature (no references).

Also pushed the cleanup: removed the now-unused `ping` / `@types/ping` from `package.json` (+ lockfile).

PR #116, opened 2026-07-02
feat: add self-service-dinein to orderMethodsToPrint

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&task=6a3d3938621e342c8bbde324)

### DIFF
--- a/src/modules/common.ts
+++ b/src/modules/common.ts
@@ -513,6 +513,11 @@ export const SERVICES: Record<string, ServiceType> = {
     label_en: 'ON THE GO',
     label_el: 'ΣΤΟ ΧΕΡΙ',
   },
+  self_service_dine_in: {
+    value: 'self_service_dine_in',
+    label_en: 'SELF SERVICE',
+    label_el: 'ΑΥΤΟΕΞΥΠΗΡΕΤΗΣΗ',
+  },
 };
 
 export const DISCOUNTTYPES: Record<string, ServiceType> = {
--- a/src/modules/settings.ts
+++ b/src/modules/settings.ts
@@ -121,6 +121,7 @@ export const PrinterSettings = z.object({
       'WOLT',
       'FAGI',
       'BOX',
+      'SELF_SERVICE_DINE_IN',
     ]),
   categoriesToNotPrint: z
     .array(z.string(), {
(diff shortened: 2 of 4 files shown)


Knorcedger (changes requested):
⚠️ Mike review complete — changes requested.

Task gate: verified Flarmio Partners task `6a3d3938621e342c8bbde324` (`quickord`, P2, type `fe_be`) and linked this PR to it.

Reviewed read-only with Claude Code `/code-review` against `main`. No builds/tests/lint/typecheck were run.

Blocking finding:
- `SELF_SERVICE_DINE_IN` is a dine-in variant, but table-number rendering still only checks `orderType === 'DINE_IN'`. If these orders carry `tableNumber` / `tableNumbers`, tickets will print the service label but omit the table, which is likely wrong operationally.
  - `src/modules/printer.ts:3174` main order print path
  - `src/modules/printer.ts:4003` order-comments print path
  - `src/modules/common.ts:1158` / `src/modules/common.ts:1175` shared receipt/order label lines

Suggested fix: introduce a small helper/set for dine-in order types, e.g. `['DINE_IN', 'SELF_SERVICE_DINE_IN'].includes(orderType)`, and reuse it in these table-number gates. If product intent is explicitly that self-service dine-in tickets should not show tables, please confirm that in the PR and I can re-review.

Non-blocking rollout note: existing printer configs still need the FE setting re-saved unless a separate migration/backfill is intended.

Knorcedger (approved):
✅ Mike re-review complete — approved.

Retracting my previous table-number blocker. Product clarification: self-service dine-in is registered/handled as a takeaway/package method, and table numbers are intentionally not wanted on these tickets. Given that, the strict `DINE_IN` table-number gates are correct and should not include `SELF_SERVICE_DINE_IN`.

Task gate remains verified: Flarmio Partners task `6a3d3938621e342c8bbde324` (`quickord`, P2, type `fe_be`) is linked to this PR.

No blocking issues found. The PR correctly adds `SELF_SERVICE_DINE_IN` to accepted printer settings/order type validation plus labels/services, matching the BE/FE enum casing.

Read-only re-review; no builds/tests/lint/typecheck were run.

PR #117, opened 2026-07-03
fix: make discount productId nullable, timeout WS connection attempts…

nina-dk:
… & install watchdog to reconnect

[Task](https://partners.flarmio.com/admin/dev/all/tasks?project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&savedView=6a0d8b33cdba6716a23271b8&task=6a476fa0b839d3ae1fd90292)


Knorcedger (approved):
Reviewed with Claude Code `/code-review` (read-only, no build/test/lint run by reviewer).

No blockers found. The WS handshake timeout + watchdog path is guarded against duplicate live sockets, `connectStartedAt` is correctly paired with each CONNECTING socket, and nullable discount `productId` is safe with existing downstream checks.

Non-blocking: in the rare case where `handshakeTimeout` itself never fires, watchdog recovery can take up to ~60s because of the 30s interval plus 30s stuck threshold. Acceptable as a backstop.

PR #120, opened 2026-07-10
fix: trim surrounding whitespace & increase contrast for better logo …

nina-dk:
…printing

[Task](https://partners.flarmio.com/admin/dev/all/tasks?status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&savedView=6a0d8b33cdba6716a23271b8&project=quickord&task=69fd2d568ccea86c7add5bb3)

### DIFF
--- a/src/modules/common.ts
+++ b/src/modules/common.ts
@@ -143,9 +143,14 @@ const saveCachedImage = (url: string, buffer: Buffer): void => {
   }
 };
 
-const downloadAndProcessImage = async (url: string): Promise<Buffer> => {
-  // Check cache first
-  const cachedImage = getCachedImage(url);
+const downloadAndProcessImage = async (
+  url: string,
+  trimTransparent = false
+): Promise<Buffer> => {
+  // Cache key includes the trim flag: the same URL processed both ways yields
+  // different rasters, so they must not share an entry.
+  const cacheKey = trimTransparent ? `${url}#trim` : url;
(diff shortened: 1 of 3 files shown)


Knorcedger (changes requested):
*Mike review complete.*

Partners task verified and linked: `69fd2d568ccea86c7add5bb3`.

Found 1 issue that should be fixed before merge:

- *P2* `src/modules/common.ts`: avoid trimming transparent padding for all markdown images. The shared `downloadAndProcessImage` path is used by markdown images, not only venue logos. If `issuerText` contains a QR/barcode image with a transparent quiet zone, flattening and trimming that transparent margin can leave code modules at the image edge and make the printed code harder or impossible to scan. Make trimming opt-in for logos, or otherwise preserve required padding for generic markdown images.

Knorcedger (approved):
*Mike rereview complete.*

Partners task verified and linked: `69fd2d568ccea86c7add5bb3`.

No blocking issues found on latest head `3816a42`.

Notes:
- The previous QR/barcode quiet-zone issue is fixed: transparent trimming is now opt-in and only `printLogo` passes it, while markdown images keep padding.
- I verified the Sharp `0.34.3` pipeline for padded transparent logos: current chain renders `300x150`, same as explicit trim-before-resize, so the logo sizing concern is not a blocker.

PR #121, opened 2026-07-14
feat: new setting for bold prices in order print

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&savedView=6a0d8b33cdba6716a23271b8&task=6a490d12041c1bcca7249b78)

### DIFF
--- a/src/modules/printer.ts
+++ b/src/modules/printer.ts
@@ -3241,6 +3241,24 @@ export const printOrder = async (
         ]);
 
         const boldOrderType = settings.textOptions?.includes('BOLD_ORDER_TYPE');
+        const boldPrices = settings.textOptions?.includes('BOLD_PRICES');
+
+        // Enlarges a money line when BOLD_PRICES is on, then restores the
+        // configured text size. Skipped while BOLD_PRODUCTS is active, since
+        // that already prints the whole product block enlarged.
+        const printBoldPrice = (print: () => void, alreadyEnlarged = false) => {
+          const enlarge = boldPrices && !alreadyEnlarged;
+          if (enlarge) {
+            printer.bold(true);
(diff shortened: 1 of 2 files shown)


Knorcedger (changes requested):
Thanks Antonina. I found one blocker before approval:

`BOLD_PRICES` is suppressed when `BOLD_PRODUCTS` is also enabled. In `src/modules/printer.ts`, product lines calculate `enlargePrice` as `boldPrices && !boldProducts`, and the shared `printBoldPrice(..., boldProducts)` calls skip the new bold-price behavior when products are already enlarged. But `BOLD_PRODUCTS` only changes text size here; it does not keep the later discount/product-total lines bold. So with both toggles enabled, the explicit `BOLD_PRICES` setting will not reliably make prices bold.

Relevant spots:
- `src/modules/printer.ts:3547` skips inline price bolding when `boldProducts` is true.
- `src/modules/printer.ts:3598` and the product-total print path pass `boldProducts` as `alreadyEnlarged`, which suppresses `printer.bold(true)` too.

Please separate the concepts: avoid double-enlarging when products are already enlarged, but still apply `printer.bold(true)` for price text when `BOLD_PRICES` is enabled.

I also noticed the implementation enlarges prices (`setTextSize(1, 0)`) in addition to bolding them. If that is intentional, fine; otherwise the task wording only asks for bold prices/order total, and double-width prices can wrap on narrower printers.

Verdict: request changes.

I did not run builds/tests/lint, per read-only review policy.

Knorcedger (approved):
Rereviewed the current diff. No actionable issues found.

PR #123, opened 2026-07-16
add venueId to fetch incident reports & increase abort timeout for lo…

nina-dk:
…ngpoll

[task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=69fdb53afa09045de57bf516&project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&task=6a58dad9eff9884407d2ee9f)


Knorcedger (approved):
Reviewed. No blocking issues found.

Non-blocking note: `AbortError` suppression is implemented inside shared `postJson`, so it applies to both `/print-jobs/poll` and `/print-jobs/result`. That looks acceptable because it only suppresses self-inflicted fetch timeouts when curl fallback succeeds, but it is slightly broader than the long-poll-specific comment implies.

Checks: I did not run builds, tests, lint, typecheck, package scripts, or CI scripts, per review constraints.

PR #124, opened 2026-07-17
fix: improve product/choices alignment when bold_products is enabled

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=69fdb53afa09045de57bf516&project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&task=6a58d232665f2e55d52a43cc)

### DIFF
--- a/src/modules/common.ts
+++ b/src/modules/common.ts
@@ -671,74 +671,10 @@ interface ProductOption {
   choices?: OptionChoice[];
 }
 
-export const printOptionDetails = (
-  printer,
-  options: ProductOption[],
-  lang: string,
-  settings: any
-) => {
-  options?.forEach((option) => {
-    let optionLabel = normalizeGreek(getTitle(option.content, lang))
-      .toUpperCase()
-      .trim();
-    if (!optionLabel.endsWith(':') && optionLabel.length > 0) {
-      optionLabel = `${optionLabel}: `;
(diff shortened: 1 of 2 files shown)


Knorcedger (changes requested):
Found one regression that should be fixed before merge.

- [P2] `src/modules/common.ts:724-726` — when a choice makes the current line exactly `width` characters, the next choice enters the wrap branch but `current.length + 1 <= width` is false, so the comma separator is not emitted before `pushCurrent()`. This regresses the previous `choiceValues.join(", ")` behavior and can print adjacent choices without a separator whenever the preceding choice exactly fills the line.

I did not run builds/tests/lint, per review-only policy.

Knorcedger (approved):
Rereviewed after Antonina’s comments and rechecked the diffs.

You were right: my previous separator finding is not approval-blocking. At the exact-width boundary the choices are separated by a line break, so they are not printed adjacent; the missing trailing comma is cosmetic/trade-off rather than ambiguous output. For the printer-server PR specifically, the flagged separator branch is also pre-existing on `main`, so it is not introduced by this PR.

No approval-blocking findings from the rereview.

I did not run builds/tests/lint, per review-only policy.

PR #125, opened 2026-07-20
fix: enhance WebSocket failure reporting & add jitter to longpoll req…

nina-dk:
…uests

[Task](https://partners.flarmio.com/admin/dev/quickord/tasks?status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&savedView=6a0d8b33cdba6716a23271b8&task=6a5dd100d054cf7d91e847f0)

### DIFF
--- a/src/modules/api.ts
+++ b/src/modules/api.ts
@@ -154,7 +154,7 @@ const reportWebSocketFailure = async (details: {
   url: string;
   venueId: string;
 }): Promise<void> => {
-  const message = `Problem: printer-server blocked from opening its WebSocket to ${details.url} for venue ${details.venueId} for ${details.outageMinutes}min (${details.attempts} attempts) — ${details.category} (${details.code}): ${details.message}. Needs a firewall/proxy/TLS fix on the venue's network; printing still works over the pull channel.`;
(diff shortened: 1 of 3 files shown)


Knorcedger (changes requested):
Requesting changes.

Blocking finding:

`src/modules/wsClient.ts:689-690` now suppresses sustained WebSocket outage reports unless the last error category is in `REPORTABLE_WS_CATEGORIES`. That excludes handshake timeouts / `UNKNOWN`, but this same file documents a key real-world blocked-WS case as a firewall/proxy accepting TCP and black-holing the upgrade. In `ws`, that commonly surfaces as an opening-handshake timeout without a useful code, so a permanently blocked venue can now retry forever with only local logs and never raise `addError`/Slack.

The new 15-minute duration gate already handles transient-noise reduction; filtering out timeout/unknown transport blocks regresses the previous “report persistent cannot-open-WS” behavior. Please keep persistent handshake timeout/unknown failures reportable, or otherwise prove those blocked cases are classified into one of the allowed categories.

The long-poll jitter change looks safe and scoped.

I did not run builds, tests, lint, or verification scripts per review instructions.

Knorcedger (approved):
Rereviewed latest changes. Task link verified against Flarmio Partners task 6a5dd100d054cf7d91e847f0.\n\nNo blocking findings. The latest commit removes the reportable-category allow-list so persistent pre-open UNKNOWN/TIMEOUT failures can now report after the sustained-outage gate, while stable/deliberate reconnect paths reset the episode state. I also checked the timer/connection cleanup paths and did not see a regression.\n\nRead-only review only: no builds/tests/lint/typecheck/install scripts were run.

PR #132, opened 2026-07-23
feat: retry failed fetch calls up to 3 times before falling back to c…

nina-dk:
…url & report results

[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=69fdb53afa09045de57bf516&project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&task=6a61d7032f3803a9b404ca82)

### DIFF
--- a/src/modules/http.ts
+++ b/src/modules/http.ts
@@ -21,6 +21,10 @@ export interface FetchFailureDetails {
   fetchErrorCode?: string;
   fetchErrorCause?: unknown;
   responseStatus?: number;
+  // How many times fetch was tried before giving up and falling back to curl.
+  // 1 with the retry loop disabled; >1 means every retry failed too — the
+  // "fetch is broken here, not a transient blip" signal.
+  fetchAttempts: number;
   curlOk: boolean;
   networkDown: boolean;
 }
@@ -28,6 +32,9 @@ export interface FetchFailureDetails {
 export interface HttpResult<T> {
   data: T;
   viaFallback: boolean;
+  // Number of fetch attempts made (whether it eventually succeeded or fell
+  // back to curl). Present on both the success and fallback paths.
+  fetchAttempts: number;
   fetchFailure?: FetchFailureDetails;
 }
 
@@ -68,12 +75,17 @@ const buildFetchErrorContext = (
 const finalize = (
   ctx: ReturnType<typeof buildFetchErrorContext>,
   fetchErr: any,
+  fetchAttempts: number,
   curlOk: boolean
 ): FetchFailureDetails => ({
   ...ctx,
+  fetchAttempts,
   curlOk,
   networkDown: isNetworkDown(fetchErr, ctx.responseStatus),
 });
+
+const sleep = (ms: number): Promise<void> =>
+  new Promise((r) => setTimeout(r, ms));
 
 // ---------- curl primitives (exec a raw curl command) ----------
 
@@ -144,6 +156,14 @@ export interface TryFetchOpts<T> {
   method: string;
   fetchFn: () => Promise<FetchFnResult<T>>;
   curlFn: () => Promise<T>;
+  // Extra fetch attempts before falling back to curl. Default 0 (single try,
+  // original behavior). When >0, a fetch failure is retried up to this many
+  // times — if a retry succeeds it was a transient blip; if all fail and curl
+  // then works, the failure is fetch-specific.
+  fetchRetries?: number;
+  // Pause between fetch attempts. Default 500ms. Only consulted when
+  // fetchRetries > 0.
+  retryDelayMs?: number;
 }
 
 // Run fetchFn; if it throws (or returns non-2xx surfaced via thrown error),
@@ -152,38 +172,59 @@ export const tryFetchWithFallback = async <T>(
   opts: TryFetchOpts<T>
 ): Promise<HttpResult<T>> => {
   const { url, method, fetchFn, curlFn } = opts;
+  const maxAttempts = 1 + Math.max(0, opts.fetchRetries ?? 0);
+  const retryDelayMs = opts.retryDelayMs ?? 500;
   let fetchErr: any;
   let responseStatus: number | undefined;
+  let attempt = 0;
 
-  try {
-    const r = await fetchFn();
-    return { data: r.data, viaFallback: false };
-  } catch (err: any) {
-    fetchErr = err;
-    responseStatus = err?.responseStatus;
+  // Try fetch up to maxAttempts times. A retry that succeeds means the earlier
+  // failure was a transient connection blip — logged (as the tracing signal)
+  // but not surfaced as a failure. Only when every attempt fails do we fall
+  // back to curl below.
+  for (attempt = 1; attempt <= maxAttempts; attempt++) {
+    try {
+      const r = await fetchFn();
+      if (attempt > 1) {
+        logger.warn(
+          { url, method, attempt, maxAttempts },
+          'fetch recovered on retry'
+        );
+      }
+      return { data: r.data, viaFallback: false, fetchAttempts: attempt };
+    } catch (err: any) {
+      fetchErr = err;
+      responseStatus = err?.responseStatus;
+      if (attempt < maxAttempts) await sleep(retryDelayMs);
+    }
   }
 
+  const fetchAttempts = maxAttempts;
   const ctx = buildFetchErrorContext(url, method, fetchErr, responseStatus);
-  logger.error(ctx, 'fetch failed, attempting curl fallback');
+  logger.error(
+    { ...ctx, fetchAttempts },
+    'fetch failed, attempting curl fallback'
+  );
 
   try {
     const data = await curlFn();
-    logger.info({ url, method }, 'curl fallback succeeded');
+    logger.info({ url, method, fetchAttempts }, 'curl fallback succeeded');
     return {
       data,
       viaFallback: true,
-      fetchFailure: finalize(ctx, fetchErr, true),
+      fetchAttempts,
+      fetchFailure: finalize(ctx, fetchErr, fetchAttempts, true),
     };
   } catch (curlErr: any) {
     logger.error(
-      { ...ctx, curlStderr: curlErr?.stderr || curlErr?.message },
+      { ...ctx, fetchAttempts, curlStderr: curlErr?.stderr || curlErr?.message },
       'curl fallback also failed'
     );
     const wrapped: any = new Error(
       `fetch and curl both failed for ${method} ${url}: ${fetchErr?.message}`
     );
     wrapped.cause = fetchErr;
-    wrapped.fetchFailure = finalize(ctx, fetchErr, false);
+    wrapped.fetchFailure = finalize(ctx, fetchErr, fetchAttempts, false);
     throw wrapped;
   }
 };
--- a/src/modules/http.ts
+++ b/src/modules/http.ts
@@ -167,6 +167,14 @@ export const isCheapRetryableFetchError = (err: any): boolean => {
   return CHEAP_RETRYABLE_CODES.has(extractErrorCode(err) ?? '');
 };
 
+// A fallback the curl path silently papers over with no action attached:
+// measured as frequent and always curl-recovered. Used to gate the result
+// report path, which can't use the poll's persistence streak. AbortError is
+// deliberately absent — on the result path's 10s timeout it's the one signal
+// that fetch is broken only there (the poll's 45s budget would survive).
+export const isRecoveredFetchNoise = (f: FetchFailureDetails): boolean =>
+  f.fetchErrorCode === 'UND_ERR_CONNECT_TIMEOUT';
+
 // ---------- curl primitives (exec a raw curl command) ----------
 
 export const curlExec = (
--- a/src/modules/http.ts
+++ b/src/modules/http.ts
@@ -22,17 +22,15 @@ export interface FetchFailureDetails {
   fetchErrorCode?: string;
   fetchErrorCause?: unknown;
   responseStatus?: number;
-  // How many times fetch was tried before giving up and falling back to curl.
-  // 1 with the retry loop disabled; >1 means every retry failed too — the
-  // "fetch is broken here, not a transient blip" signal.
+  // >1 means every retry failed too: fetch is broken here, not blipping.
   fetchAttempts: number;
   curlOk: boolean;
   networkDown: boolean;
   // A curl that answers fast where fetch hung means the link was fine.
   fetchDurationMs?: number;
   curlDurationMs?: number;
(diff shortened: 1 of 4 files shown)


Knorcedger (changes requested):
Found one blocking issue:

- [P2] Keep poll retries off result reporting (`src/modules/pullClient.ts:133-134`)

`postJson()` is shared by both `/print-jobs/poll` and `/print-jobs/result`, so adding `fetchRetries: POLL_FETCH_RETRIES` there makes result reporting try fetch up to three times before curl as well. With `RESULT_TIMEOUT_MS` (10s), a single result-report attempt can now take roughly 31s before the existing retry delays. In the fetch-hanging scenario this PR targets, that can push an acknowledgement for a successfully printed job past the backend's claim-result timeout, causing false failure/lost-job reporting. Please make the retry behavior path-specific, for example apply the extra retries to long-poll only and keep result reporting on the faster fetch→curl path.

Task gate verified: Flarmio Partners task `6a61d7032f3803a9b404ca82` is Quickord BE and linked to this PR.

Note: Codex accidentally attempted `npm test -- --runInBand` despite my read-only/no-verification instruction; it failed immediately because the repo has no `lint` script. No build/test/lint/typecheck verification was used for this verdict.

Knorcedger (changes requested):
I re-reviewed `fc94659` read-only and found one remaining issue.

[P2] Preserve noise suppression for non-retried result reports (`src/modules/pullClient.ts:187-199`)

`postJson` is shared by both `/print-jobs/poll` and `/print-jobs/result`. The new retries are only enabled for the poll call, but the old `recoveredNoise` suppression was removed for the whole shared helper. For `/print-jobs/result`, `fetchRetries` still defaults to `0`, so an `AbortError` or `UND_ERR_CONNECT_TIMEOUT` recovered immediately by curl is still the same single-attempt transient noise the old code intentionally suppressed. This will start surfacing known benign result-report blips as hourly incidents, even though only the poll path has the extra attempts that justify treating fallback as persistent.

Knorcedger (changes requested):
Re-reviewed latest head `7f37542` read-only. The previous result-reporting retry/noise-suppression issue is fixed, but I found one remaining blocking concern:

- [P2] Avoid delaying curl fallback behind repeated long-poll timeouts (`src/modules/pullClient.ts:47`, `src/modules/pullClient.ts:267`)

`POLL_FETCH_RETRIES = 2` means the long-poll path can now wait behind up to three full `POLL_TIMEOUT_MS` fetch attempts before trying curl. In the fetch-broken-but-curl-ok venue failure mode this PR targets, where fetch hangs until `AbortError`, curl now runs after roughly 135s instead of after the first 45s timeout, so the pull channel can delay print-job delivery for 2+ minutes per poll cycle even though curl would have worked. Result reporting correctly avoids this because of its claim-result deadline, but poll delivery still has user-facing latency. Consider falling back to curl immediately for known timeout/connect-timeout classes, or using a much shorter retry timeout before curl.

Non-blocking note: `fetch recovered on retry` is logged as an unthrottled warning from `src/modules/http.ts:188`; if noisy venue links recover via retry often, consider demoting/throttling that log so it doesn't replace the incident noise the PR is trying to reduce.

Task gate: PR still contains the Flarmio Partners task link.

Knorcedger (changes requested):
Re-reviewed latest head `62370c8` read-only (isolated worktree, base `main`). No build/test/lint/typecheck was run.

The previous round's blocking concern is **fixed**: `isCheapRetryableFetchError` excludes `AbortError`, so the long poll's own 45s timeout now goes straight to curl instead of re-holding the poll twice more, and `fetch recovered on retry` was demoted to `info`. The retry-attempt accounting in `tryFetchWithFallback` is also correct in both exit paths (`break` on non-retryable, and exhausting `maxAttempts`), and the now-required `fetchAttempts` field doesn't break any of the six `tryFetchWithFallback` call sites.

Two blocking issues remain, both in the new diagnostics rather than the retry logic itself.

---

### [P2] `socketReused` is permanently `undefined` for any URL with a query string (`src/modules/http.ts:69`, `src/modules/http.ts:87`)

The write key and the read key don't agree:

- setter (`sendHeaders`): `` `${request.method} ${request.path}` `` — undici's `request.path` is **pathname + search**
- reader (`requestTraceKey`): `` `${method} ${new URL(url).pathname}` `` — **pathname only**

I confirmed this on Node 24 with a read-only probe: for `GET https://api.github.com/repos/.../releases/latest?per_page=1&foo=bar`, undici publishes `path = "/repos/.../releases/latest?per_page=1&foo=bar"`. The `pathname` lookup can never hit it.

The consequence is worse than a missing field, because `FetchFailureDetails.socketReused` documents `undefined` as *"never got as far as sending headers (connect-phase failure)"*. So for `common.ts`'s image fetches and the two `autoupdate.ts` calls — all of which use `redirect: 'follow'` and land on query-bearing signed CDN/GitHub URLs — every incident will assert a connect-phase failure regardless of what actually happened. That's the opposite of the disambiguation this field exists to provide.

Two smaller things fall out of the same key scheme:

- The key drops the **host**, so two origins sharing a pathname (`/`, `/releases/latest`) overwrite each other's mark.
- `socketReuseByRequest` is a plain `Map` fed by *every* undici request in the process once tracing starts, and it's only ever `delete`d for keys `tryFetchWithFallback` itself looks up. Unique signed-URL paths from image and release downloads accumulate entries that are never reclaimed. Slow, but it's a long-lived service.

Suggested fix: use `pathname + search` on the read side and include the origin on both sides, e.g. `` `${method} ${u.host}${u.pathname}${u.search}` `` derived from the same URL parse.

For what it's worth, the underlying `WeakSet` + consuming `delete()` reuse detection is sound — I verified it reports `reused=false` on a fresh connection and `reused=true` on the third sequential request over a kept-alive socket. Only the keying is wrong.

### [P2] Concurrent result reports can push the fallback streak to 3 from a single brief blip (`src/modules/pullClient.ts:179-180`, `src/modules/pullClient.ts:104-110`)

The constant's comment sets out the intent precisely: *"No single event tells them apart, only persistence — so count consecutive fallbacks."* But `consecutiveFetchFallbacks` has no time component and is shared by both paths `postJson` serves, and `/print-jobs/result` calls are genuinely concurrent — `dispatchJob` hands each job to `executePrintJob`, which serializes per *printer*, so an order fanning out to kitchen + bar + receipt printers produces three detached `reportResult` calls in flight at once.

If the uplink hiccups for ~2s while those three are in flight, all three fall back to curl, the counter reaches `FETCH_FALLBACK_STREAK_TO_REPORT`, and an incident fires. That's one instant, not persistence — and it's the single-attempt result-report noise round 2 of this review asked to keep suppressed, arriving through the new gate instead. Note that the result path still runs `fetchRetries: 0`, so each of those three contributes a streak point off a *single* failed attempt, while a poll contributes one off three.

The mirror-image case also exists, though it's narrower: on a busy venue, successful result reports reset the counter between poll cycles, so a middlebox that kills only long-held connections (fetch broken for the poll, fine for short POSTs) may never accumulate three in a row.

Suggested fix: track the streak per path (`poll` vs `result`) rather than one module-level counter, and/or require the streak to span a minimum wall-clock window so concurrent in-flight requests can't satisfy it in one instant.

---

### Non-blocking

- **Retry-on-5xx re-amplifies the thundering herd the jitter guards against** (`src/modules/http.ts:158`, `src/modules/pullClient.ts:66-72`). `isCheapRetryableFetchError` returns true for any `status >= 500`. `POLL_ERROR_BACKOFF_JITTER`'s comment explains that a dyno restart fails every venue's held long poll in the same millisecond, and that a fixed wait would march them all back onto the freshly-booted dyno together. With `POLL_FETCH_RETRIES = 2`, each venue now sends 3 fetch attempts + 1 curl within ~1s of that restart instead of 1 + 1, un-jittered, before the jittered backoff ever applies. Partly self-limiting (`ECONNREFUSED` isn't retryable, so it breaks immediately), but it does double the burst at the moment the file says the dyno is least able to absorb one. Consider excluding 503 specifically, or jittering `FETCH_RETRY_DELAY_MS`.
- **No overall deadline across attempts** (`src/modules/http.ts:261-282`). Each attempt gets a *fresh* `POLL_TIMEOUT_MS` AbortController, so the retry budget is bounded only by how fast a retryable error happens to surface. The `AbortError` exclusion closes the common 45s case, but a retryable code surfacing late (e.g. `ETIMEDOUT` at 40s) still costs ~2min before curl. A `Date.now()` budget checked before each retry would make the "these attempts cost seconds, not minutes" claim in `POLL_FETCH_RETRIES`'s comment structurally true rather than incidentally true.
- **`fetchDurationMs` is last-attempt-only, not total** (`src/modules/http.ts:278`). It's reassigned each iteration, so on a 3-attempt failure the incident reports one attempt's duration next to `fetchAttempts: 3`. Reasonable as the like-for-like comparison against `curlDurationMs`, but the field name reads as total — worth a word in the comment or a rename.
- **The PR adds a comment-length rule it then breaks** (`CLAUDE.md:101-103`). The new Code Style section says *"Keep comments short. One or two lines... No multi-paragraph rationale blocks above constants or functions."* The same diff adds a 10-line rationale block above `POLL_FETCH_RETRIES` (`src/modules/pullClient.ts:46-55`) and a 5-line one above `FETCH_FALLBACK_STREAK_TO_REPORT`. I'd rather keep the rationale than lose it — this codebase's comments are unusually good — but the rule as written should be softened to match, or it'll read as dead advice from day one.

No test suggestions: the repo's only test is `test/printer.test.ts`, and there's no existing pattern covering `http.ts` or the pull loop.

---

**Task gate:** ✅ satisfied — the PR body links Flarmio Partners task `6a61d7032f3803a9b404ca82`.

**Verdict:** changes requested. Both P2s are small, contained fixes (a key expression and where the streak counter lives); the retry design itself now looks right.

src/modules/http.ts
```diff
@@ -39,6 +54,55 @@ const extractErrorCode = (err: any): string | undefined => {
...
+  try {
+    diagnosticsChannel.subscribe('undici:client:connected', (msg: any) => {
+      if (msg?.socket) freshSockets.add(msg.socket);
+    });
+    diagnosticsChannel.subscribe('undici:client:sendHeaders', (msg: any) => {
+      const request = msg?.request;
+      if (!request?.method || !request?.path) return;
+      const key = `${request.method} ${request.origin}${request.path}`;
+      // Skip redirect targets and any request we aren't awaiting a mark for.
+      if (!inFlightTraceKeys.has(key)) return;
```
Knorcedger: **P2 — Consume the socket's fresh mark even for ignored redirect hops.** Returning here before `freshSockets.delete(msg.socket)` leaves a newly connected redirect-target socket marked fresh. If undici later reuses that socket for a directly tracked request, this subscriber deletes the stale mark and records `socketReused: false`, even though the socket was reused—the exact opposite of the diagnostic's meaning. I reproduced this with a redirect followed by two direct requests to the target origin: the failing request used the redirect's same remote port, while `FetchFailureDetails.socketReused` was `false`. Please consume `freshSockets` for every `sendHeaders` event, then gate only the `socketReuseByRequest.set(...)` call on `inFlightTraceKeys`; that keeps redirect keys out of the Map without preserving stale freshness.

Knorcedger (changes requested):
The prior redirect-target Map leak is bounded at this head, and the retry/result-reporting behavior checks out. One P2 remains in the replacement guard: ignored redirect hops do not consume the socket's fresh mark, so a later request can report a genuinely reused socket as fresh and misdirect the incident diagnosis this field was added for.

Verification at exact head: typecheck passed; the complete Jest suite passed (1/1); the production build passed; source formatting and `git diff --check` passed. These local checks ran on Node 26 because the declared Node 24.8 runtime is not installed here.

src/modules/http.ts
nina-dk: fixed in [fef976b](https://github.com/Knorcedger/quickord-printer-server/pull/132/commits/fef976bd006ea36bf726b38f9e3d63db09d2c08b).

src/modules/pullClient.ts
```diff
@@ -149,6 +185,16 @@ async function postJson(
     url,
   });
 
+  if (trackFallbackStreak) {
+    // Only transport failures count. A 401/404/5xx came back over a connection
+    // fetch opened fine, so curl "recovering" it says nothing about fetch —
+    // and a bad-creds venue would otherwise sit at the threshold forever.
+    if (!result.viaFallback) consecutiveFetchFallbacks = 0;
+    else if (result.fetchFailure?.responseStatus === undefined) {
```
Knorcedger: [P2] Reset the streak when fetch receives an HTTP response

This branch deliberately excludes 401/404/5xx fallbacks from the transport-failure streak, but it leaves the previous streak value untouched. For example: two transport fallbacks, then a fetch 503 that curl recovers, then one transport fallback reports `consecutiveFallbacks: 3`. Those transport failures were not consecutive—the intervening response proves fetch opened a connection—so the diagnostic can still page on non-persistent events. Please reset `consecutiveFetchFallbacks` whenever `responseStatus` is defined (the 401 path throws just below this block anyway), and add a sequence test if this helper is made testable.

Knorcedger (changes requested):
The previous socket-freshness finding is fixed at exact head `fef976b`: untracked redirect hops now consume the fresh-socket mark before the trace-key guard. One P2 remains in the new fallback-streak filter: an HTTP-response fallback is excluded from the counter but does not reset an existing streak, so non-consecutive transport failures can still reach the incident threshold. See the inline comment for the sequence and contained fix.

Verification at this head: typecheck passed; the complete Jest suite passed (1/1); the production build passed; `git diff --check` passed. Checks used Node 26 because the declared Node 24.8 runtime is not installed. The aggregate `npm test` command still fails before Jest on both base and PR because `package.json` references a nonexistent `lint` script; I ran the underlying checks separately.

Knorcedger (approved):
The latest change fixes the remaining fallback-streak issue: fetch success, any HTTP response fallback, and a curl-also-failed uplink now all reset the poll-only streak, while only curl-recovered transport fallbacks advance it. I rechecked the surrounding poll/result paths and found no remaining blocking issues.

Verified at exact head 5139ab38: typecheck passed; complete Jest suite passed (1/1); production build passed; source/diff checks passed. Checks used Node 26 because the declared Node 24.8 runtime is not installed here. The aggregate `npm test` command remains independently broken on both base and PR because it references a nonexistent `lint` script, so its real checks were run separately.

src/modules/pullClient.ts
nina-dk: [5139ab3](https://github.com/Knorcedger/quickord-printer-server/pull/132/commits/5139ab382ff16598ee3463e604cac28cf0b3f649)

PR #133, opened 2026-07-29
feat: remove WS path

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&savedView=6a0d8b33cdba6716a23271b8&assignee=69fd2d4b8ccea86c7add5aae&task=6a60852927ddfc699cae7b05)


Knorcedger (approved):
Approving — clean removal, extraction is byte-faithful, full message-type parity with the deleted WS handler. Four stale WS comments noted in the review comment above (non-blocking), plus one cross-repo item to confirm before merge: that the BE derives PS liveness from the poll rather than an open socket.

PR #135, opened 2026-08-05
feat: add multi-modem support per venue

nina-dk:
[Task]( https://partners.flarmio.com/admin/dev/all/tasks?project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&savedView=6a0d8b33cdba6716a23271b8&task=6a5e14da0cfed7d30f33b5e7)

### DIFF
--- a/docs/modem.md
+++ b/docs/modem.md
@@ -4,6 +4,7 @@
   - [Drivers](#drivers)
   - [Modem AT Commands](#modem-at-commands)
   - [Setup](#setup)
+  - [Multiple modems (one per phone line)](#multiple-modems-one-per-phone-line)
 
 ## Drivers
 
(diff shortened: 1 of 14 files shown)


Knorcedger (approved):
Re-review of `334df4c`: all three findings from the previous round are fixed and each landed with a test that pins the behaviour (parser RING poisoning, superseded-instance open race, legacy `modem` mirror carrying `venueId`).

Approving. Five low-severity/nit items and a note on test wiring are in the review status comment — none blocking:

1. `closeInstance` doesn't await a superseded instance's in-flight `open()`, so a settings POST landing during modem startup collides on the COM port and recovers via the 1s backoff (`modem.ts:298-307`).
2. Keepalive interval isn't cleared when reconnect permanently gives up (`modem.ts:200-205`).
3. `modemId` is dead schema surface — never written or read anywhere (`settings.ts:227`).
4. `no-await-in-loop` disable added for one loop but not the one above it; `no-underscore-dangle` (error in `.eslintrc`) flags the new `__` test hooks. Doesn't break CI since no `lint` script exists.
5. Nit: `shouldEmit` is consumed before the `venueId` guard, so a dropped call still burns the 5s dedup slot.

Separate follow-up worth a task: the 600+ lines of tests added here aren't run by anything — the only workflow triggers on push to `main` and runs `deploy.sh`, and `npm test` calls a `lint` script that doesn't exist (both pre-existing). A PR job running `npm run jest` would make this coverage protective.

Knorcedger (approved):
Re-reviewed at `61573cf`. Both round-2 fixes verified (keepalive teardown on give-up, dedup after the venueId guard), each with an assertion. Two new findings — a pre-existing parser bug around withheld caller ID, and a latent legacy-modem fallback trap — both non-blocking and detailed in the review status comment.

PR #137, opened 2026-08-18
ci: pin windows-2025 runner & verbose nexe build logs

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&task=6a7f1b02e2554fe2b902b85a)

## Why

The [main-branch release run for #135](https://github.com/Knorcedger/quickord-printer-server/actions/runs/31795491107) failed with `vcbuild.bat nosign release x64 exited with code: 1`.

Root cause (two things compounding):

1. The `nexe-Windows-24.8.0` Actions cache was evicted after 8 quiet days (GitHub deletes caches unused for 7 days), forcing the first build-Node-from-source run in months — every retained run since May had restored the cached binary and never exercised this path.
2. Meanwhile (between the Jun 9 and Jun 23 runs) GitHub rolled the `windows-latest` alias from the `windows-2025` image (VS 2022) to `windows-2025-vs2026`. Node 24's `vcbuild.bat` can't detect VS 2026, so the source build dies ~2s after source extraction, at toolchain detection.

## Changes

- **Pin `runs-on: windows-2025`** — the image that still ships VS 2022, which `vcbuild.bat` can find. Alias rolls of `windows-latest` can no longer silently swap the compiler under this workflow.
- **Add `--verbose` to the nexe build** in `build:exe` so vcbuild's actual output lands in CI logs instead of a bare exit code.

## Notes

- The repo's Actions cache is currently completely empty, so the first run after merge will do the full ~30–40 min Node source build, then repopulate the cache; subsequent runs return to normal speed.
- Merging this triggers a release run (as any main push does).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #138, opened 2026-08-18
ci: move to windows-2022 runner (only image left with VS 2022)

nina-dk:
[Task](https://partners.flarmio.com/admin/dev/all/tasks?savedView=6a0d8b33cdba6716a23271b8&project=quickord&status=todo%2Cin_progress%2Cin_review%2Cpaused%2Cdone&assignee=69fd2d4b8ccea86c7add5aae&task=6a7f1b02e2554fe2b902b85a)

## Why the pin in #137 didn't work

The Aug 18 run ([32109210650](https://github.com/Knorcedger/quickord-printer-server/actions/runs/32109210650)) requested `windows-2025` but the job log shows GitHub served the **`windows-2025-vs2026`** image anyway — GitHub has since migrated the `windows-2025` label itself onto the VS 2026 image. Per [actions/runner-images](https://github.com/actions/runner-images#available-images), `windows-latest`, `windows-2025`, and `windows-2025-vs2026` are now all the same VS-2026-only image.

The `--verbose` flag from #137 did surface the real error this time:

```
Using ClangCL because the Node.js version being compiled is >= 24.
Looking for Visual Studio 2022
Failed to find a suitable Visual Studio installation or Clang compiler/LLVM toolset.
```

Node 24.8.0's `vcbuild.bat` searches for VS in version range `[17.6,18.0)` — VS 2022 only. VS 2026 (18.x) support was only added to `vcbuild.bat` in Node **24.14.0** (Feb 2026).

## Fix

Use `windows-2022` — the last hosted image that ships VS 2022 (Enterprise 17.14, including the `VC.Llvm.Clang` + `ClangToolset` components the Node ≥24 build requires). This is the same VS generation that successfully built the cached 24.8.0 binary back in Sept 2025.

## Caveats

- First post-merge run still builds Node from source (~30–40 min) since the nexe cache is empty; the cache is keyed on `runner.os` (`Windows`), so it carries over across image changes.
- `windows-2022` is the oldest Windows image and will eventually be retired. The durable fix is bumping the nexe target to Node ≥24.14 (VS 2026-aware vcbuild) — left out of this PR deliberately since it changes the shipped runtime and needs its own testing.

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -10,8 +10,8 @@ permissions:
 
 jobs:
   create:
-    # Pinned: Node 24's vcbuild.bat can't detect VS 2026, which windows-latest now ships.
-    runs-on: windows-2025
+    # Node 24.8's vcbuild.bat needs VS 2022; windows-2025/latest now ship only VS 2026.
+    runs-on: windows-2022
 
     steps:
       # Checkout repo


Knorcedger:
## 🔍 Code review — in progress

### Timeline
- 08:19  Review started
- 08:21  Checked out PR head `c5dbcdf` in an isolated worktree (base `main`); 1 changed file — `.github/workflows/windows.yaml` (+2/-2)
- 08:22  Confirmed task link present in PR body ✅ · pulled workflow run history (last green Aug 6, failures Aug 14 + Aug 18) · repo Actions cache is currently empty (0 entries)
- 08:23  Verified against `actions/runner-images`: `windows-2022` is still **GA**, and `windows-latest` / `windows-2025` / `windows-2025-vs2026` are indeed the same VS-2026 image — the PR's premise checks out
- 08:26  Read the failed run log (32109210650) — reproduces the PR's quoted `Failed to find a suitable Visual Studio installation` verbatim; last green run was only 6m23s, i.e. it restored the nexe cache rather than building Node, so cache eviction is what exposed the broken build path
- 08:27  Cross-checking `deploy.sh` / `build:exe` nexe target and the rest of the job's toolchain steps against the 2022 image

